PR #6, opened 2018-04-10
fix: import { configure } statement, and notes…

travomic:
Calling out path to `.storybook/config.js` file, and updating the import statement to match latest `@storybook/react` path.


tmeasday:
Thanks so much @travomic!

PR #7, opened 2018-04-11
Fix import source of { createTask, actions }

travomic:
The './Task.stories' import source is no longer valid, and should come from: './Task.story'


tmeasday:
Thanks @travomic! Actually I think we got it backwards, in the code https://github.com/hichroma/learnstorybook-code/blob/master/.storybook/config.js it is `.stories.js`. I will change the text as that is easier (29e2220).

Thanks for the PR!

PR #8, opened 2018-04-11
fix: Source of { withPinnedTasks } import

travomic:
Sorry for the teeny individual change sets, but this is the same issue as the prior PR, I assume the naming convention used to be *.stories.js


tmeasday:
Moving back to `.stories.js` thanks again!

PR #13, opened 2018-04-12
Refactor props

dburles:
This avoids passing in dynamic props which are a problem for snapshot testing


tmeasday:
This is great, thanks @dburles.

I'll make the corresponding changes in the source code before I merge this.

PR #14, opened 2018-04-12
#add: get-started ES

icarlossz:
I begin the translation of the first element called get-started.

### DIFF
--- a/content/spanish/get-started.md
+++ b/content/spanish/get-started.md
@@ -5,54 +5,54 @@ description: "Setup React Storybook in your development environment"
 commit: 30939d5
 ---
 
-# Get started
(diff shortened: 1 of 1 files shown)


tmeasday:
Awesome! I think you should also translate the frontmatter (`title` / `tocTitle` / `description`) so it comes up translated in the menu etc?

I'll give you commit access and you can commit / coordinate with others (@luchux) directly?

icarlossz:
Oh, you're right, I just saw it in the guide, in the next commit I'll fix it. :)

luchux:
I translated them in the revision ;)

ready to be merged in next PR.

PR #23, opened 2018-05-03
Update simple-component.md

ng-hai:
I saw the result didn't match with the explanation video.


domyen:
Thanks for suggesting this Hai. I appreciate you taking the time to make a PR. 🙏

In this case the omission of `${state}` is intentional.  We want to show a simple version of the component at the early part of the chapter and slowly graduate to more complexity. You'll see that `${state}` is added later in the "Build out the states" section. I'll close this for now.

ng-hai:
Yeah, I know the intention, just feel confusing when comparing the result with the video on ARCHIVED state.

PR #34, opened 2018-06-18
Made a start on Vue text

tmeasday:
Corresponding commits: https://github.com/hichroma/learnstorybook-code/commits/vue

### Needed to downgrade from babel@7 

It looks like the current vue version is on babel@7, storybook does not yet support it. There's probably a case that we wait for SB@4 which will support babel@7 before releasing this tutorial?

### Needed to add a `webpack.config` and remove `getstorybook`-generated `.babelrc`

I think `getstorybook` should do both of these things.

The `webpack.config` (which I mostly copied from @mattrothenberg) adds support for `.vue` files and vue's CSS handling. This is AFAICT pretty standard vue stuff and should probably just always be done.

## Questions so far

I made some changes around how components are registered and actions passed through the stack. 

Need to check in with Vue people to see if they are weird or not. They fit better with the flow of the text from the React version.


domyen:
LGTM, reviewed and merging now. Thanks @mattrothenberg and @tmeasday!

PR #37, opened 2018-06-30
fix(data): restore missing prop on PureTaskList

coogie:
The `PureTaskList` function is missing the `loading` prop in its destructuring


tmeasday:
Thankyou @coogie !

PR #38, opened 2018-06-30
fix(data): make Jest test consistent

coogie:
The Jest test code snippet doesn't really match the [previous chapter](https://github.com/hichroma/learnstorybook.com/blob/master/content/react/en/composite-component.md#unit-tests-with-jest)'s, just bringing it in line :)


tmeasday:
Much appreciated! This highlights that the code was wrong, fixing!

PR #41, opened 2018-07-16
yarn test - watchman

nTamura:
Can no longer chown /usr/local on high sierra 

https://github.com/Homebrew/brew/issues/3228

sudo chown -R $(whoami) $(brew --prefix)/*

### DIFF
--- a/content/react/en/get-started.md
+++ b/content/react/en/get-started.md
@@ -37,8 +37,8 @@ yarn run storybook
 yarn start
 ```
 <div class="aside">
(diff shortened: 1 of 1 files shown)


tmeasday:
Hey @nTamura, thanks for contributing!

I think maybe this information is best added to the CRA advice about installing watchman? 

Perhaps we should link directly to : https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-test-hangs-on-macos-sierra -- and you could maybe send a PR to that file and add some more info. I think it is way more likely to stay up to date with things than our little note here ;)

PR #44, opened 2018-07-23
Add support for Traditional Chinese

danielduan:
I'm not sure if we'd like to use either simplified or traditional as the source of truth and set up a script to automatically convert between the two. There might be some idiomatic difference between the two languages that scripts like OpenCC may not fully account for.

Hope this can help users outside of mainland China.


danielduan:
I was going to fix some of the consistency issues around formatting and missing image alt tag translations in this PR, but it's actually a lot more than I expected.

I will do that in a following PR when I have time.

domyen:
One last thing: could you add a link to traditional chinese in the view layer and language picker in `Header.js`?

Everything else looks awesome!

PR #46, opened 2018-07-27
Fix deploy settings for @netlify

flexbox:
There is an error during the build time if we don't add a `.` on the folder name

image


tmeasday:
Hey @flexbox -- is this supposed to be a `png`?

flexbox:
Hi @tmeasday 
In the netlify section https://www.learnstorybook.com/react/en/deploy/
there is a small typo on the picture the folder name needs to be `.storybook-static/` instead of `storybook-static/`

kylemh:
screen shot 2018-07-29 at 11 04 56 pm

@tmeasday it wasn't an accident. GitHub hides the image, but here's the change as @flexbox describes.

tmeasday:
Thanks for catching me up there guys, I was a bit slow.

I'm confused though. I'm pretty sure `storybook-static` *is* correct and is the default output directory of `build-storybook`. I wonder what's going on?

kylemh:
According to the directions here: https://www.learnstorybook.com/react/en/deploy/

`build-storybook` outputs to `.storybook-static`

flexbox:
@tmeasday 
If the image is the source of truth, we need to modify the documentation as @kylemh said.
Modify the script like this
`build-storybook` needs to output to `storybook-static`

I had this issue when I was deploying to netlify
07_30_18 at 11 58 01am

I just had a look to the storybook documentation.
They says to use `.out` folder.
https://storybook.js.org/basics/exporting-storybook/#docs-content

I think it's better for concistency to keep the `.` for folder name

@tmeasday Let me know if you need more informations, because I am confused as well 😆

tmeasday:
Ah, ok that explains it!

I'm not sure, it looks like the rest of the text in the deploy chapter assumes that the output will be in `storybook-static` (and that is the default), so I am going to go ahead and just remove the `-o` flag from the command for now (https://github.com/chromaui/learnstorybook.com/commit/16a56ca3be3e26c43bd08c4999035d999ddf1906)

I think it makes sense for this tutorial to just use the default. I'm not sure why storybook's docs recommend not using the default? If it is better to use `.out` or something else dotted, then that should be the default I would have thought. Either way I think this tutorial should follow SB's lead.

tmeasday:
Thanks so much for bearing with me on this one @flexbox and @kylemh! Closing this as I think the problem is solved now.

Issue #2, opened 2018-04-03
Angular code


domyen:
Angular users, we need your help converting the [LearnStorybook example code](https://github.com/hichroma/learnstorybook-code). 🙏 

The example is currently written in React and Redux. If you use Angular (and a little React) and have a grasp on its state management ecosystem help us introduce Storybook to that community.

Comment below!

alterx:
Hey @domyen, I helped bringing Angular to Storybook and would love to contribute here. I've also done a ton of React work so it shouldn't be an issue to convert the code :) 

I don't have a lot of free time but will start looking into the code. Is there anything I should know before starting?

tmeasday:
Hey @alterx, great to hear from you!

I'm glad you asked. I've been (slowly) looking at trying to get the Vue version integrated as contributed by @mattrothenberg broken up into steps to match our chapters (see 
https://github.com/chromaui/learnstorybook.com/issues/1#issuecomment-397947421)

The thing that takes a little while is just getting every step working in a simple reproducable way for users. In particular the first chapter where you run a few commands and end up with working app and storybook and jest environment---this wasn't quite as easy in Vue as it was in React (at least for me, who isn't a Vue user).

So basically what I am trying to say is there are a few levels of work that need to be done:

1. Getting the final app&storybook as per https://github.com/chromaui/learnstorybook-code running in an angular environment w/ equivalent code (e.g. replacing Redux w/ something that makes sense for angular, etc). This bit I definitely can't do myself and is the most important!

2. Getting the code from above broken up into commits that match up with the chapters and are easy for the user to follow with simple steps. I appreciate this is a pain and I am happy to do it myself although I suspect you will be better at it than me!

3. Writing the corresponding text to 2. This is easier to do at the same time as 2, but also I am more than happy to do it.

Does that make sense? Please hit me up here or on storybook slack if there are any specific questions!

alterx:
Hey @tmeasday, long time no see (chat 😂) !

I like the idea of having a commit per title, it feels like it's going to help a lot. As of the work that needs to be done, I intend to work on the three parts. It's probably easier (and just faster) for the person that wrote the app to work on 2 and 3.

I'll start with the app and hit you up on Slack if I have further questions :)

alterx:
The code is here https://github.com/chromaui/learnstorybook-code/pull/4
Will also add changes for the site on this repo.

domyen:
Thank you @alterx! 🙏 Closing this ticket

PR #47, opened 2018-07-27
Added initial Angular content

alterx:
This PR accompanies https://github.com/chromaui/learnstorybook-code/pull/4 and addreses #2 

What's inside:
- [x] English version

What's missing:

- [x] Spanish version
- [ ] Update commit SHAs for all the pages after that PR gets merged
- [ ] Proofread the whole thing (I could use some help with this one 😄)
- [ ] 🚀

### DIFF
--- a/content/angular/en/composite-component.md
+++ b/content/angular/en/composite-component.md
@@ -1,0 +1,288 @@
+---
+title: "Assemble a composite component"
+tocTitle: "Composite component"
+description: "Assemble a composite component out of simpler components"
+commit: '5776042'
+---
+
+# Assemble a composite component
+
+Last chapter we built our first component; this chapter extends what we learned to build TaskListComponent, a list of Tasks. Let’s combine components together and see what happens when more complexity is introduced.
+
+## TasklistComponent
+
+Taskbox emphasizes pinned tasks by positioning them above default tasks. This yields two variations of `TaskListComponent` you need to create stories for: default items and default and pinned items.
+
+![default and pinned tasks](/tasklist-states-1.png)
+
+Since `TaskComponent` data can be sent asynchronously, we **also** need a loading state to render in the absence of a connection. In addition, an empty state is required when there are no tasks.
+
+![empty and loading tasks](/tasklist-states-2.png)
+
+## Get setup
+
+A composite component isn’t much different than the basic components it contains. Create a `TaskListComponent` component and an accompanying story file: `src/tasks/task-list.component.ts` and `src/tasks/task-list.stories.ts`.
+
+Start with a rough implementation of the `TaskListComponent`. You’ll need to import the `TaskComponent` component from earlier and pass in the attributes and actions as inputs and events.
+
+```typescript
+import { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
+import { Task } from './task.model';
+
+@Component({
+  selector: 'task-list',
+  template: `
+    <div class="list-items">
+      <task-item
+        *ngFor="let task of tasksInOrder"
+        [task]="task"
+        (onArchiveTask)="onArchiveTask.emit($event)"
+        (onPinTask)="onPinTask.emit($event)"
+      >
+      </task-item>
+    </div>
+  `,
+})
+export class TaskListComponent implements OnInit {
+  tasks: Task[] = [];
+  @Input() loading: boolean = false;
+  @Output() onPinTask: EventEmitter<any> = new EventEmitter();
+  @Output() onArchiveTask: EventEmitter<any> = new EventEmitter();
+
+  constructor() {}
+
+  ngOnInit() {}
+}
+
+```
+
+Next create `Tasklist`’s test states in the story file.
+
+```typescript
+import { storiesOf, moduleMetadata } from '@storybook/angular';
+import { CommonModule } from '@angular/common';
+
+import { TaskComponent } from './task.component';
+import { TaskListComponent } from './task-list.component';
+import { task, actions } from './task.stories';
+
+export const defaultTasks = [
+  { ...task, id: '1', title: 'Task 1' },
+  { ...task, id: '2', title: 'Task 2' },
+  { ...task, id: '3', title: 'Task 3' },
+  { ...task, id: '4', title: 'Task 4' },
+  { ...task, id: '5', title: 'Task 5' },
+  { ...task, id: '6', title: 'Task 6' },
+];
+
+export const withPinnedTasks = [
+  ...defaultTasks.slice(0, 5),
+  { id: '6', title: 'Task 6 (pinned)', state: 'TASK_PINNED' },
+];
+
+const props = {
+  tasks: defaultTasks,
+  onPinTask: actions.onPinTask,
+  onArchiveTask: actions.onArchiveTask,
+};
+
+storiesOf('TaskList', module)
+  .addDecorator(
+    moduleMetadata({
+      declarations: [TaskListComponent, TaskComponent],
+      imports: [CommonModule],
+    }),
+  )
+  .add('default', () => {
+    return {
+      template: `
+        <div style="padding: 3rem">
+          <task-list [tasks]="tasks" (onPinTask)="onPinTask($event)" (onArchiveTask)="onArchiveTask($event)"></task-list>
+        </div>
+      `,
+      props,
+    };
+  })
+  .add('withPinnedTasks', () => {
+    return {
+      template: `
+        <div style="padding: 3rem">
+          <task-list [tasks]="tasks" (onPinTask)="onPinTask($event)" (onArchiveTask)="onArchiveTask($event)"></task-list>
+        </div>
+      `,
+      props: {
+        ...props,
+        tasks: withPinnedTasks,
+      },
+    };
+  })
+  .add('loading', () => {
+    return {
+      template: `
+        <div style="padding: 3rem">
+          <task-list [tasks]="[]" loading="true" (onPinTask)="onPinTask($event)" (onArchiveTask)="onArchiveTask($event)"></task-list>
+        </div>
+      `,
+      props,
+    };
+  })
+  .add('empty', () => {
+    return {
+      template: `
+        <div style="padding: 3rem">
+          <task-list [tasks]="[]" (onPinTask)="onPinTask($event)" (onArchiveTask)="onArchiveTask($event)"></task-list>
+        </div>
+      `,
+      props,
+    };
+  });
+
+```
+
+`addDecorator()` allows us to add some “context” to the rendering of each task. In this case we add the module metadata so we can use all the Angular components inside out stories.
+
+<div class="aside">
+<a href="https://storybook.js.org/addons/introduction/#1-decorators"><b>Decorators</b></a> are a way to provide arbitrary wrappers to stories. In this case we’re using a decorator to add metadata.
+</div>
+
+`task` supplies the shape of a `Task` that we created and exported from the `task.stories.ts` file. Similarly, `actions` defines the actions (mocked callbacks) that a `TaskComponent` expects, which the `TaskListComponent` also needs.
+
+Now check Storybook for the new `TaskList` stories.
+
+<video autoPlay muted playsInline loop>
+  <source
+    src="/inprogress-tasklist-states.mp4"
+    type="video/mp4"
+  />
+</video>
+
+## Build out the states
+
+Our component is still rough but now we have an idea of the stories to work toward. You might be thinking that the `.list-items` wrapper is overly simplistic. You're right – in most cases we wouldn’t create a new component just to add a wrapper. But the **real complexity** of `TaskListComponent` is revealed in the edge cases `withPinnedTasks`, `loading`, and `empty`.
+
+```javascript
+iimport { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
+import { Task } from './task.model';
+
+@Component({
+  selector: 'task-list',
+  template: `
+    <div class="list-items">
+      <task-item
+        *ngFor="let task of tasksInOrder"
+        [task]="task"
+        (onArchiveTask)="onArchiveTask.emit($event)"
+        (onPinTask)="onPinTask.emit($event)"
+      >
+      </task-item>
+
+      <div *ngIf="tasksInOrder.length === 0 && !loading" class="wrapper-message">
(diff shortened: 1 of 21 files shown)


content/angular/en/simple-component.md
```diff
@@ -0,0 +1,385 @@
...
+  title: string;
+  state: string;
+}
+```
+
+## Component built!
+
+We’ve now successfully built out a component without needing a server or running the entire frontend application. The next step is to build out the remaining Taskbox components one by one in a similar fashion.
+
+As you can see, getting started building components in in isolation is easy and fast. We can expect to produce a higher-quality UI with less bugs and more polish because it’s possible to dig in and test every possible state.
```
domyen: Typo: Double "in" here

content/angular/en/composite-component.md
```diff
@@ -0,0 +1,288 @@
+---
+title: "Assemble a composite component"
+tocTitle: "Composite component"
+description: "Assemble a composite component out of simpler components"
+commit: '5776042'
+---
+
+# Assemble a composite component
+
+Last chapter we built our first component; this chapter extends what we learned to build TaskListComponent, a list of Tasks. Let’s combine components together and see what happens when more complexity is introduced.
```
domyen: Change to "a list of TaskComponents"

content/angular/en/composite-component.md
```diff
@@ -0,0 +1,288 @@
...
+    type="video/mp4"
+  />
+</video>
+
+## Build out the states
+
+Our component is still rough but now we have an idea of the stories to work toward. You might be thinking that the `.list-items` wrapper is overly simplistic. You're right – in most cases we wouldn’t create a new component just to add a wrapper. But the **real complexity** of `TaskListComponent` is revealed in the edge cases `withPinnedTasks`, `loading`, and `empty`.
+
+```javascript
+iimport { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
```
domyen: iimport --> import

content/angular/en/composite-component.md
```diff
@@ -0,0 +1,288 @@
...
+  });
+});
+
+```
+
+![TaskList test runner](/tasklist-testrunner.png)
+
+Note that we’ve been able to reuse the `withPinnedTasks` list of tasks in both story and unit test; in this way we can continue to leverage an existing resource (the examples that represent interesting configurations of a component) in more and more ways.
+
+Notice as well that this test is quite brittle. It's possible that as the project matures, and the exact implementation of the `Task` changes --perhaps using a different classname or a `textarea` rather than an `input`--the test will fail, and need to be updated. This is not necessarily a problem, but rather an indication to be careful liberally using unit tests for UI. They're not easy to maintain. Instead rely on visual, snapshot, and visual regression (see [testing chapter](/test/)) tests where possible.
```
domyen: `Task` --> `TaskComponent`

content/angular/en/screen.md
```diff
@@ -0,0 +1,192 @@
...
+      template: `<inbox-screen></inbox-screen>`,
+    };
+  })
+  .add('error', () => {
+    return {
+      template: `<inbox-screen></inbox-screen>`,
+    };
+  });
+```
+We see that although the `default` story works just fine (because we have default data in our store), we have an issue in the `error` story, because the `error` field is connected to the store itself and we can't just modify it from outside. We need to dispatch and action on the store.
```
domyen: "...dispatch and action"  ---> "...dispatch an action"

content/angular/en/test.md
```diff
@@ -0,0 +1,153 @@
...
+Add the package as a dependency.
+
+```bash
+yarn add storybook-chromatic
+```
+
+Import Chromatic in your `.storybook/config.js` file.
+
+```javascript
+import { configure } from '@storybook/react';
```
domyen: /react ---> /angular

domyen:
Also, could you update the links in `Header.js` to link to the English (and Spanish) editions. 🙏

alterx:
@icarlossz, well, I've already started with that one. As soon as I get a first draft I'll submit it and you can proofread it. Does that work? 😄

icarlossz:
@alterx It sounds great to me!

alterx:
@icarlossz I've added the Spanish version
@domyen I've addressed your change requests (and a couple more I found)

content/angular/en/test.md
```diff
@@ -66,12 +66,14 @@ yarn add storybook-chromatic
 Import Chromatic in your `.storybook/config.js` file.
 
 ```javascript
-import { configure } from '@storybook/react';
+import {
+  configure
+} from '@storybook/angular';
 import 'storybook-chromatic/storybook-addon';
+import '../src/styles.less';
```
domyen: This one should remain .css. The user is not actually running less, rather they're including the pre-compiled .css file included in the repo.

alterx: Well, I changed that in both the code and in the tutorial. I've configured the cli to use LESS. It didn't really make sense to use gulp and compile less when the @angular/cli + webpack can allow use of .less files.

alterx:
Hey @tmeasday, regarding the `PureTaskList` thing, I'm not really sure it makes sense for Angular. Let me try to rationalize it: 

In React, when we use `connect` we compose a new higher order component that will provide the "selected" data to the underlying "presentational" component. Both components exist on their own and `connect` just enhances what the presentational component can do.

In Angular, in order to create a "connected" component we just inject the store and then we can select the data inside. It felt a bit unnecessary to add that part since we already have a presentational component (`TaskList`) and we'll build a "container" (`InboxScreenComponent`). So, if I had included that `PureTaskList` the `TaskList` would've basically been the same thing as the `InboxScreenComponent` and we weren't going to be able to just import "the presentational part" of the component. 

In other words, there's no similar way to have the `PureTaskList` and `TaskList` in the same file and connect one while leaving the other one as a pure component like React allows. The only way would be to have two different component definitions: one for the dummy `PureTaskList` and one for the "connected" `TaskList` that would basically include a `PureTaskList` and a reference to the store, which is exactly how the relation between `InboxScreenComponent` and `TaskList` looks like.

Because of that, it felt weird to add it. It's basically replicating the same thing twice. I don't know if this is clear enough, tho. Let me know if you get what I'm trying to say.

I do agree that the `data` section feels a bit out of place on its own, and I actually thought about unifying both parts but didn't want to make a change that big to the structure of the lessons.

tmeasday:
@alterx -- thanks for the write up, that makes perfect sense to me.

I feel like the issue involved here is simply that it isn't convenient to define two components in the one file in Angular as it is in React. Otherwise everything you've said above applies to React as well, and I struggled as well with the fact that `InboxScreen` doesn't really do much above the containerised `TaskList`. 

Here's what I was thinking and I'll let you decide if it is worth the awkwardness of defining *another* component:

 - It is a common pattern to include container components deep in the hierarchy of your application (as opposed to only at the top-level, cf `InboxScreen`).
 - A useful pattern to allow you easily write stories for container components is to export/split out the presentational component inside and target the stories at that. (This is the lesson that we are trying to impart in the data chapter).
 - This works great at the level of that container, but you cannot apply that technique to any components higher up the tree because they include the container. 
 - A useful pattern to deal with writing stories for "container-containing" components is to mock out the data source. (This is the lesson that we are trying to impart in the screens chapter).

Although in this example app it really doesn't make a lot of sense to have the container not at the top-level, in order to try and demonstrate both of the above I made things a bit convoluted and showed *errors* at the screen level, but fetch the *data* at the `TaskList` level.

I realise you wouldn't write an app like this in reality but things are typically way more complex and you are often tempted to embed containers deep in the hierarchy. So these techniques are good to know!

(As we mention in the screen chapter it is possible to avoid embedded containers completely in your app and not ever do the data-mocking technique, but that's not a good fit for a lot of (esp. redux) apps).

---

I wonder: if splitting the presentational part from the data access in a container is unnatural in Angular, is there a way to do the first technique in some other way? I feel like it's a good idea for people to know about!

alterx:
Exactly!, I was struggling to see the need for the two components even in the React app but now that you explain why I get it and it makes sense. The main difference is that, in Angular, that separation of components isn't as "convenient" and felt even more awkward without knowing the context.

The way of doing this in Angular is actually creating both components. I think I could add that and clarify a bit the message 😄

tmeasday:
Awesome! 

I think maybe I need to work on the text a bit to make the above reasoning a bit more obvious so people aren't thinking "this is weird why would I ever do this" ;)

content/angular/en/test.md
domyen: sgtm thanks for the explanation 👍

alterx:
@tmeasday 

So, after a bit of thinking, I'm still not sure this part is really useful in the Angular app. In Angular, as in React, we do have `connected` components all over the component tree (not only at the top level) but the component definition mechanism and the separation between dumb-smart components works a bit different and the way components are bootstrapped removes some of the issues this section tries to solve.

> Here's what I was thinking and I'll let you decide if it is worth the awkwardness of defining another component:
> 
> - It is a common pattern to include container components deep in the hierarchy of your application (as opposed to only at the top-level, e.g. InboxScreen).
> - A useful pattern to allow you easily write stories for container components is to export/split out the presentational component inside and target the stories at that. (This is the lesson that we are trying to impart in the data chapter).
> - This works great at the level of that container, but you cannot apply that technique to any components higher up the tree because they include the container. 
> - A useful pattern to deal with writing stories for "container-containing" components is to mock out the data source. (This is the lesson that we are trying to impart in the screens chapter).

1) is also true in Angular.
2) this is the one I have more issues with since it doesn't work/feel the same way in Angular. When we talk about "presentational components" in Angular we just have a plain old component that only has `@Input()` and `@Output()` fields. It doesn't perform any other tasks outside of processing the inputs, rendering and, if necessary, sending out events. A connected component just gets a store via dependency injection and does it's thing, that's the main difference. So, splitting the presentational component out basically means leaving the `TaskListComponent` as is and create a new "connected" component that renders `TaskListComponent` and provides it with the task list. 
3) with Angular that isn't a problem: as long as you've injected the right thing via the `providers` array and the component is included in the ngModule you'll be able to render all the things, I've created an example in a separate branch: https://github.com/alterx/learnstorybook-code/commit/74df8a2794ccf5186c6512c383f43a6c9724dfd7
As you can see, other than the `error` related code, there's not much difference between the `InboxScreenComponent` and the `ConnectedTaskListComponent` (and their `.stories.ts` files) and rendering the stories for `InboxScreenComponent` even though it includes the `ConnectedTaskListComponent` is trivial as long as the right`providers` and components are available.
4) In this case we "mock" it by injecting the real thing and it "Just Works™" (again, this is due to Angular's DI system). 

After this explanation and reviewing the code I've added to that branch do you feel like it makes sense to add it to this PR? I just can't shake the feeling that we'd just be repeating ourselves here

tmeasday:
Hey @alterx thanks again for bearing with me and explaining things to me, I appreciate it.

I'm still not sure things are so different between Angular and React--notice that we could also achieve an equivalent thing to your `ConnectedTaskListComponent`'s story in React/Redux with something like (linked):

```js
  // This Angular:
  .addDecorator(
     moduleMetadata({
       declarations: [],
       imports: [TaskModule, NgxsModule.forRoot([TasksState])],
       providers: [Store],
     }),
   )

  // Is equivalent to this React:
  import { Provider } from 'react-redux';
  import store from '../lib/redux';

  .addDecorator((story) => (
     <Provider store={store}>{story()}</Provider>
  );
```

(I think anyway, apologies if I am getting this wrong).

So, again, I'm not sure things are all that different, I think I still just haven't explained why I did it this way well enough ;). Let me have another try:

1. Note that we *don't* write a story for the connected component in learn storybook, so in some ways the awkwardness of `ConnectedTaskList` and `InboxScreen` being the same is not so apparent 😊 

2. Not that in the code snippet above, we are just letting the default state of the store render. I'm not sure that makes a good story -- really we should be providing starting state as part of the story so it is clearer what's going on. I'm sure that's possible but it's a little added complexity for the "injecting/context" approach.

3. In the case of more complex tools (or real redux stores), it may be a *lot* more complex. For instance, I am used to Apollo/GraphQL, where again it is possible to mock the results of a query for a `Query` component, but it is much more brittle than providing the props directly and definitely a order of magnitude more complex than the "split into connected/presentational, write stories for presentational" approach.

For this reason, I really think it's good to show people that technique, definitely for React where the splitting is pretty easy and natural and probably advisable for other reasons too. For Angular, where it is awkward, I get why you aren't keen to do it. 

I'm not sure. I still think the technique may be useful for Angular people, and I would suggest we show it to them even if it doesn't really make sense for this particular app. But I am by no means an Angular expert (or even user) so I will completely defer to your instincts on this. Let me know what you want to do!

alterx:
Yeah, I think my main issue is that we're basically creating two components that are fundamentally  the same thing (`TaskListComponent` becomes `PureTaskListComponent`, then we create a new `ConnectedTaskListComponent` then, in the next chapter, we create an `InboxScreenComponent` that is almost identical to the ConnectedTaskListComponent. It feels like duplicating things for the sake of showing a complex pattern that may not be apparent to a person that's new to the framework and/or Storybook. So it boils down to the example being so contrived and my mind feeling like these components are just too much haha, but then again, I know that we should intentionally keep things simple.

I'll add the new components.

alterx:
@tmeasday , I've added the new text in english, please take a look and tell me what you think. 

I had to change the "We see that although the `error` story works just fine, we have an issue in the `default` story, because the `TaskList` has no Redux store to connect to. (You also would encounter similar problems when trying to test the `PureInboxScreen` with a unit test)." part in screen.md because in Angular the fact that the `PureInboxScreenComponent` includes the `TaskListComponent` requires all the dependencies of the latter to also be included. In other words, `PureInboxScreenComponent` is not really pure so even if I submit the `error` prop I also need to submit all the other properties and providers (the store, for example).

This is one of the things I thought wouldn't make a lot of sense to add (the 'PureInboxScreenComponent`) but I added it for the sake of completeness when comparing this to its React counterpart. 

Please let me know if the text makes sense and conveys what you wanted to convey in the React text so I can translate the missing parts to spanish.

alterx:
I think this last commit solves the whole thing while showing an alternate way of defining stories :)

tmeasday:
I think this is great @alterx! Let's get this merged!

tmeasday:
We probably still need to fix up the commit links .. I can do in the morning..

PR #49, opened 2018-08-24
Minor changes to Vue/en tutorial pages content

jarobar435:
df79878  
With the previous configuration the snapshots were not generated.  
The related [commit (b2274bd)](https://github.com/hichroma/learnstorybook-code/commit/b2274bd) shows proper path and file name.

f866397  
Invalid import.

### DIFF
--- a/content/vue/en/simple-component.md
+++ b/content/vue/en/simple-component.md
@@ -225,7 +225,7 @@ With the [Storyshots addon](https://github.com/storybooks/storybook/tree/master/
 yarn add --dev @storybook/addon-storyshots jest-vue-preprocessor
 ```
 
-Then create an `src/storybook.test.js` file with the following in it:
+Then create an `tests/unit/storybook.spec.js` file with the following in it:
 
 ```javascript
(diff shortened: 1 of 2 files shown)


content/vue/en/simple-component.md
```diff
@@ -225,7 +225,7 @@ With the [Storyshots addon](https://github.com/storybooks/storybook/tree/master/
 yarn add --dev @storybook/addon-storyshots jest-vue-preprocessor
 ```
 
-Then create an `src/storybook.test.js` file with the following in it:
+Then create an `tests/unit/storybook.spec.js` file with the following in it:
```
icarlossz: Wouldn't changing the direction of the file and folders make it necessary to change it where it is used?

jarobar435: Yes, it is necessary for the tests to work.
It's just a leftover from React guide that was modified in the code itself (check the mentioned commit), but was left in tutorial content due to inattention.

There are other files yet to be updated in same way (like test content from vue/en/data - another React belongings). Will fix them on my way throughout the guide.

PR #53, opened 2018-10-12
Update get-started.md

travis5491811:
Added note that yarn error might also be from yarn not being installed and a link to official repo.


tmeasday:
Thanks @travis5491811!

PR #57, opened 2018-10-23
Fix grammar in vue setup guide

samit4me:


tmeasday:
Thanks very much @samit4me !

PR #58, opened 2018-10-23
Get vue styles working

samit4me:
Updated the "Reuse CSS" section of the vue docs to show how to reuse the CSS inside a vue app. It is also possible to import the CSS into `main.js` but if you did that it's likely you would also need to remove the default styles in `App.vue` as it would override what's in `index.css`. So I've updated to what I believe is the easiest solution.

### DIFF
--- a/content/vue/en/get-started.md
+++ b/content/vue/en/get-started.md
@@ -82,7 +82,13 @@ Depending on what part of the app you’re working on, you may want to run one o
 
 ## Reuse CSS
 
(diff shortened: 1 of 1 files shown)


samit4me:
I just realised there is a "Keep your code in sync" section at the bottom pointing to https://github.com/chromaui/learnstorybook-code/commit/d1c4858 which probably needs updating also as it doesn't appear to be importing the styles either. I'm not 100% sure how to go about that at the moment. Does anyone have some ideas?

tmeasday:
Thanks @samit4me! I'll update the source and merge this soon, probably as part of a general update to SB4 in the next day or two -- once it is released!)

(The source that needs to be changed is here: https://github.com/chromaui/learnstorybook-code btw, but it's slightly complicated changing it as you need to rewrite the git history)

samit4me:
Sounds awesome @tmeasday, thank you for that 😃

tmeasday:
Sorry this took so long @samit4me !

PR #70, opened 2019-01-11
Various fixes to the Vue english tutorial

alejandroiglesias:
I've just followed the tutorial and found some issues/inconsistencies that I fixed in this PR. Please let me know if some of them need clarification.

### DIFF
--- a/content/vue/en/deploy.md
+++ b/content/vue/en/deploy.md
@@ -21,7 +21,7 @@ To deploy Storybook we first need to export it as a static web app. This functio
 ```javascript
 {
   "scripts": {
-    "storybook": "build-storybook -c .storybook -o .storybook-static"
+    "build-storybook": "build-storybook -c .storybook -s public -o storybook-static"
   }
 }
 ```
--- a/content/vue/en/deploy.md
+++ b/content/vue/en/deploy.md
@@ -26,7 +26,7 @@ To deploy Storybook we first need to export it as a static web app. This functio
 }
 ```
 
-Now when you run Storybook via `npm run storybook`, it will output a static Storybook in the `storybook-static` directory.
+Now when you run Storybook via `npm run build-storybook`, it will output a static Storybook in the `storybook-static` directory.
 
 ## Continuous deploy
 
--- a/content/vue/en/screen.md
+++ b/content/vue/en/screen.md
@@ -79,7 +79,7 @@ export default {
 </script>
 ```
 
-We also change the `App` component to render the `InboxScreen` component (eventually we would use a router to choose the correct screen, but let's not worry about that here):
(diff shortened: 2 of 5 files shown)


content/vue/en/screen.md
```diff
@@ -79,7 +79,7 @@ export default {
 </script>
 ```
 
-We also change the `App` component to render the `InboxScreen` (eventually we would use a router to choose the correct screen, but let's not worry about that here):
+We also change the `App` component to render the `InboxScreen` component (eventually we would use a router to choose the correct screen, but let's not worry about that here):
```
icarlossz: I don't know if it's necessary to add this word because the name of the component itself indicates what it's referring to.

content/vue/en/deploy.md
```diff
@@ -21,7 +21,7 @@ To deploy Storybook we first need to export it as a static web app. This functio
 ```javascript
 {
   "scripts": {
-    "storybook": "build-storybook -c .storybook -o .storybook-static"
+    "build-storybook": "build-storybook -c .storybook -s public -o storybook-static"
```
icarlossz: The command should be `npm run storybook` or `yarn storybook` so this is fine only with `storybook`.
```suggestion
    "storybook": "build-storybook -c .storybook -s public -o storybook-static"
```

icarlossz:
Hi @alejandroiglesias , I left you some comments.

content/vue/en/screen.md
alejandroiglesias: Then it may be better to say `to render InboxScreen` instead of `to render the InboxScreen`. Otherwise, the sentence looks a bit strange to me (disclaimer: I'm not a native english speaker).

alejandroiglesias:
@icarlossz thanks for the review, added some replies.

content/vue/en/screen.md
icarlossz: `to render InboxScreen` makes sense.

alejandroiglesias: Actually, I found a lot of sentences to like that (`to render the X`) through many files, so I've reverted this change in order to avoid changing a lot of files.

content/vue/en/deploy.md
icarlossz: so, if this change here it's necessary to change the script name in the following paragraph as well

> Now when you run Storybook via `npm run **build-**storybook`, it will output a static Storybook in the `storybook-static` directory.

alejandroiglesias: Good catch. Done.

content/vue/en/screen.md
icarlossz: 👍

tmeasday:
Will defer to you on this on @icarlossz

icarlossz:
Merged, thank you Alejandro

PR #71, opened 2019-01-13
add redux: Provider & store to storybook

andihaki:
fixing issue:

Could not find "store" in the context of "Connect(PureTaskList)". Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(PureTaskList) in connect options.

### DIFF
--- a/content/react/en/data.md
+++ b/content/react/en/data.md
@@ -120,9 +120,11 @@ However, we can easily solve this problem by simply rendering the `PureTaskList`
 ```javascript
 import React from 'react';
 import { storiesOf } from '@storybook/react';
+import { Provider } from "react-redux";
 
 import { PureTaskList } from './TaskList';
 import { task, actions } from './Task.stories';
+import store from "../lib/redux";
 
 export const defaultTasks = [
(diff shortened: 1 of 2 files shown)


tmeasday:
Hi @andihaki - I think perhaps you missed the bit where we export the original (non-redux) `TaskList` as `PureTaskList` and then the connected one as the default export?

andihaki:
Hi @tmeasday - already did what u said in my code. but still get error until i add this 3 line of codes: 
```javascript
import { Provider } from "react-redux";
import store from "../lib/redux";
.addDecorator(story => <Provider store={store}><div style={{ padding: '3rem' }}>{story()}</div></Provider>)
```
if you think everything is fine, then just closed my PR 👍

tmeasday:
@andihaki is it possible you wrote:

```js
import PureTaskList from './TaskList';
```

rather than
```js
import { PureTaskList } from './TaskList';
```

andihaki:
wow, you're right. it was my bad

PR #85, opened 2019-04-02
add redux provider section to the data.md

mostafa69d:
The provider from the react-redux is not mentioned in the docs and make the App and test cases fail. Added the steps to add the provider to the App


mostafa69d:
I can see it's been added in the next section, `screens`, maybe it's better to add

PR #89, opened 2019-05-09
Replace storybook.test.js with storyshots.test.js

luiskhernandez:
In latest version, using storybook.test.js throws an error


domyen:
@tmeasday do we need to update the code example for this one?

tmeasday:
What error is thrown? I ran through this just the other day and didn't see a problem

domyen:
Seems inactive so closing this to make maintenance more efficient. Thanks for raising though!

PR #94, opened 2019-05-23
Fix linting errors

kylesuss:
Quite a few linting errors right now. This clears all of them up. Would be good eventually to get CI going for this repo to check this stuff in PRs, but that can maybe come a bit later 😉 

I clicked around on the site after adding the prop-types to see if I could see any places that were expecting something as required but not receiving it. I didn't see any!


.eslintrc.js
```diff
@@ -3,9 +3,20 @@ module.exports = {
   extends: ['airbnb', 'prettier', 'plugin:jest/recommended'],
   plugins: ['prettier', 'jest'],
   env: {
+    'browser': true,
```
kylesuss: So you can reference `window`, `document`

.eslintrc.js
```diff
@@ -3,9 +3,20 @@ module.exports = {
...
+    'graphql': 'readonly',
+  },
   rules: {
+    'jsx-a11y/anchor-is-valid': [
+      'error',
+      {
+        'components': ['GatsbyLink'],
+        'specialLink': ['to'],
+      },
+    ],
```
kylesuss: Since `GatsbyLink`s don't have an `href` prop

package.json
```diff
@@ -30,7 +30,7 @@
     "build": "gatsby build",
     "develop": "gatsby develop",
     "dev": "yarn develop",
-    "format": "prettier-eslint --write 'src/**/*.js'",
+    "format": "prettier-eslint --write '**/*.js'",
```
kylesuss: I figure it makes sense to lint config files too. I added a `.eslintignore` file as well to allow for ignoring directories that sit in the root

src/layouts/index.js
```diff
@@ -22,7 +22,11 @@ const HeaderWrapper = styled(Header)`
 `;
 
 const TemplateWrapper = ({
-  data: { site: { siteMetadata: { title, permalink, description, githubUrl, toc } } },
+  data: {
+    site: {
+      siteMetadata: { title, permalink, description, githubUrl, toc },
+    },
+  },
```
kylesuss: Not sure why prettier didn't catch this in my last PR 🤔

PR #95, opened 2019-05-23
Upgrade gatsby from v1 to v2

kylesuss:
# Heads up that the base branch for this is `master`. I talked to @domyen about merging this into a branch called `next`, but that branch doesn't generate deploy previews via netlify (and I don't have netlify access). Therefore, I left the base as `master` so that someone can review this and validate that it works on the preview, but ultimately this should merge into another branch called `next` which will be a work-in-progress branch for a lot of upcoming work.

The API changed a bit from v1 to v2 of Gatsby. I worked through this lil guide for the process:

https://www.gatsbyjs.org/docs/migrating-from-v1-to-v2/

I also updated all the Gatsby related libs to their newest versions. However, there is a plugin that I am not sure if it will work after the upgrade:

https://github.com/stephenmathieson/gatsby-plugin-segment

There is a chance it works fine and we just need to swap it out if we want to do anything more advanced. Any help/thoughts on how to troubleshoot whether or not the old plugin still works though?


jonniebigodes:
@kylesuss i was already working with the upgrade, but nice that you are already commiting the update. That plugin in question might offer some issues, i haven't fully tested it. But there's always a alternative if it does not work out of the box. If either @tmeasday or @domyen doesn't go over it and merges i don't mind going over the pull request in more detail in the days to come. Sounds good?

kylesuss:
@jonniebigodes sounds good, thanks for the reply. I just left a note in the PR description about the base branch -- I had talked to @domyen about not merging this into master just yet (it should instead merge into a branch called `next`). That said, I can't validate it on Netlify without having `master` be the base branch sadly.

Regarding the segment plugin, I just gave it a go on the netlify preview. Looks good to me. I see the Segment request and payload:

Screen Shot 2019-05-23 at 4 40 23 PM

Followed by the response:

Screen Shot 2019-05-23 at 4 40 16 PM

jonniebigodes:
@kylesuss no need to thank, glad i'm able to help out.  I was checking the code on that plugin a couple of days ago to see if it was in accord with the api as is, this because the since the that version of the peer dependency of gatsby and today alot has changed. But with your test it's looking good, so one less problem. 👍

domyen:
LGTM!

PR #98, opened 2019-06-03
Fix export to match commit on sample github repo

calaway:
Following this instruction actually broke my Task component. I went to the sample app on GitHub and copied their implementation and it fixed it. This now matches the implementation in the sample app, as seen in this commit:

https://github.com/chromaui/learnstorybook-code/commit/ea50fb1bd511be5dced6626eda887398a1b2bee2


tmeasday:
Thanks @calaway!

PR #99, opened 2019-06-05
Add Storybook design system, remove local duplications

kylesuss:
This PR integrates [`@storybookjs/design-system`](https://github.com/storybookjs/design-system) into the app. Some APIs have changed so there was some work put into reworking them. I've also removed all of the duplicate style/component files (except for `Highlight`, which is [awaiting PR review](https://github.com/storybookjs/design-system)) in this repo since the design system is now the _Source of Truth_ ™ .

Next step after this is to start cruising on the new design 💯 

This base for this PR is the `next` branch. It shouldn't deploy to the live site as some things are sort of broken, aka:

Screen Shot 2019-06-04 at 6 27 00 PM

...but they aren't worth fixing yet as they aren't in the new design. The `next` branch will serve as the working branch for the new design and the broken pieces will eventually be phased out in an upcoming PR.


.eslintignore
```diff
@@ -1,2 +1,3 @@
 .cache/**/*.js
 public/**/*.js
+src/html.js
```
kylesuss: This file is copied from Gatsby, so I think it is in our best interest to keep it as they format it for ease of update in the future

package.json
```diff
@@ -29,7 +29,6 @@
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
-    "dev": "yarn develop",
```
kylesuss: Dup of `yarn develop`

src/html.js
```diff
@@ -0,0 +1,38 @@
+// File contents copied from Gatsby's cache in order to add a link tag for the font
+// Docs: https://www.gatsbyjs.org/docs/custom-html/
```
kylesuss: I noticed that the layout component was re-rendering a bit. That shouldn't happen, but I couldn't figure out why -- the only real issue with it was that the Google font stylesheet kept getting reimported because the styled-components `GlobalStyle` was re-rendering. This caused the fonts to flicker on every page change which is annoying.

This file is a lil workaround for that -- it allows us to import the fonts using a standard `link` tag. Now the flicker is gone 👌 

I documented the reasoning for the file at the top as well as provided a link to the docs on Gatsby's site.

package.json
tmeasday: `yarn dev` is what we use in other projects so maybe we should get rid of `develop`? Of course that's there because that's what Gatsby puts down by default. Happy for you to make the call here..

src/components/Header.js
```diff
@@ -248,27 +203,15 @@ export default function Header({ githubUrl, inverse, framework, firstChapter, is
...
+                          <LanguageLink to={`/vue/en/${firstChapter}/`}>English</LanguageLink>
+                          <LanguageLink to={`/vue/pt/${firstChapter}/`}>Português</LanguageLink>
                         </Detail>
                       </Meta>
                     </TooltipItem>
                   </TooltipList>
                 }
               >
-                <NavLink className={inverse ? 'inverse' : 'tertiary'}>
+                <NavLink {...(inverse ? { inverse: true } : { tertiary: true })}>
```
tmeasday: Would this read better as `<NavLink inverse={inverse} tertiary={!inverse}>`?

tmeasday:
Do we run Chromatic on this? We should?

package.json
kylesuss: I'm not too picky on this -- just think it makes sense to only have 1. Since `dev` is used in other projects and it is shorter, let's just go w/ that. I'll update.

src/components/Header.js
kylesuss: Yeah good call. That's cleaner. I'll update.

kylesuss:
@tmeasday Re: Chromatic, there are no stories yet! I am going to be getting stories & storybook going in the next PR so that might be the right time to get Chromatic going as well.

PR #100, opened 2019-06-05
Add storybook, new homepage, guide hierarchy, new chapter designs

kylesuss:
Merging into the `next` branch & then going to reopen so that the deploy preview exists....


src/__page_stories__/index.stories.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
kylesuss: Notice that the directory is weird here ☝️  -- Gatsby is weird about putting stories in the `pages` directory, so you have to put them somewhere else. They suggest making a dir just for the page stories on the guide for Storybook:

https://www.gatsbyjs.org/docs/visual-testing-with-storybook/

PR #102, opened 2019-06-10
Update composite-component / grammar edit

ryan-stephen:
Changed tense "has" in line 219.
Changed punctuation to "," in line 219.
Added "," in line 223.
Changed "more and more" to "many" in line 246.
Added missing word "about" in line 248.


domyen:
Nice thank you

PR #108, opened 2019-06-16
[PT] Angular translation for get-started.md

norato:


jonniebigodes:
@norato looks like that one slipped through the radar. Good catch on it.

PR #125, opened 2019-08-08
fix comment line that falls of canvas

rimian:
The end of this comment can't be seen in the container. This makes it a little shorter without changing the meaning.

### DIFF
--- a/content/react/en/data.md
+++ b/content/react/en/data.md
@@ -38,7 +38,7 @@ export const actions = {
   PIN_TASK: 'PIN_TASK',
 };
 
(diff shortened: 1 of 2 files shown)


kylesuss:
Nice. Thanks @rimian . Do you think you could update this file w/ the same change as well?

https://github.com/chromaui/learnstorybook.com/blob/e8405dccc3ce5db143bc4dc3f22ec0b092b22e30/content/react/pt/data.md

rimian:
Hi @kylesuss. For sure! Done already :)

PR #126, opened 2019-08-08
Missing semi colons

rimian:
Missing from `taskStateReducer`. I did all the languages.


rimian:
oops that other commit shouldn't be there.

PR #128, opened 2019-08-16
Error in simple-component.md

rich-TIA:
Incorrect path in import statement in Task.stories.js - makes SB look for Task in the current directory

### DIFF
--- a/content/react/en/simple-component.md
+++ b/content/react/en/simple-component.md
@@ -53,7 +53,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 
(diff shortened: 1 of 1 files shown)


kylesuss:
Hi @rich-TIA ! I just took a look and notice that at the top of the code block, there is this line:

`// src/components/Task.stories.js`

Given that structure,

`./Task` seems like the correct way to import it. Unless something else is wrong?

rich-TIA:
You're right, I missed that and got it to work in /src/stories because the RE still worked. Apologies, I'll withdraw the pull request.

kylesuss:
No worries @rich-TIA ! Thanks for the PR.

PR #132, opened 2019-08-22
Update the phrase "less bugs" to "fewer bugs"

openjck:
The word "fewer" may be a better choice in this situation.
dictionary.com provides a good overview of the topic:

https://www.dictionary.com/e/fewer-vs-less/


domyen:
Nice - good catch 🙏

PR #133, opened 2019-08-22
Use a colon on the "Simple component" page

openjck:
A colon may be a better choice here, especially because the second
sentence is incomplete on its own. grammarbook.com provides a good
overview of this topic:

https://www.grammarbook.com/punctuation/colons.asp


domyen:
Appreciate that @openjck! 🙇‍♂️

PR #134, opened 2019-09-10
Update composite-component.md

andrestone:
At "Data requirements and props" section, the following statement is neither present or implied with "...":
`import Task from './Task';`

### DIFF
--- a/content/react/en/composite-component.md
+++ b/content/react/en/composite-component.md
@@ -193,6 +193,8 @@ As the component grows, so too do input requirements. Define the prop requiremen
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import Task from './Task';
+
 function TaskList() {
   ...
 }


kylesuss:
Thanks @andrestone! Just took a look and I think the intention was to only include the imports which were necessary for the specific code block that you are looking at. It does seem that `import Task from './Task'` is shown in the code snippets where the `Task` component is actually used.

With that in mind, do you think it is confusing to show the import and not have it used anywhere in the code snippet?

andrestone:
Hi @kylesuss! Two observations:
a) If the intention was to only include imports which were necessary for the specific code block, using the "..." element for consistency would make much more sense, since it's used sometimes for the same purpose.
b) In this particular case, using the "..." element to represent the other imports that be in the code at that point of the Tutorial isn't the appropriate solution, notably because `Task` component is indeed used in the code snippet:
```
...
TaskList.propTypes = {
  loading: PropTypes.bool,
  tasks: PropTypes.arrayOf(Task.propTypes.task).isRequired, //usage of the Task component
  onPinTask: PropTypes.func.isRequired,
  onArchiveTask: PropTypes.func.isRequired,
};
...
```
Thanks for reviewing.

PR #135, opened 2019-09-11
Emphasizing a key modification to TaskList.js

andrestone:
It's neither trivial or standard practice to export both "presentational component" (`PureTaskList`) and Redux "connected component", but current version silently does so.


tmeasday:
Fair enough, thanks for this @andrestone

PR #136, opened 2019-09-20
Fix misspelled

saeedreza:


jonniebigodes:
looks like that one slipped through the net, nice catch nonetheless 👍

PR #144, opened 2019-09-30
Add customized logic for Twitter sharing by guide

kylesuss:
New logic to customize the Twitter sharing message by the guide. If a guide does not have the `twitterShareText` in the guide frontmatter, the Twitter share link will not be shown. It is done that way right now to account for the Visual Testing Handbook which does not yet have content so it felt weird to encourage people to share it yet.

I decided to do `twitterShareText` rather than `twitterShareUrl` because the URL encoding makes having the whole URL in the frontmatter quite messy. The URL that we end up including in the tweet is auto-generated based on the guide you are on (so we link directly to the guide), but you can customize the messaging that comes before that.

### DIFF
--- a/content/design-systems-for-developers/index.md
+++ b/content/design-systems-for-developers/index.md
@@ -62,6 +62,7 @@ contributors: [
       detail: "UX Development at Shopify",
     },
 ]
+twitterShareText: "I’m learning about building design systems! They're great for scaling frontend code on large teams."
 ---
 
 <h2>What you'll build</h2>
--- a/content/intro-to-storybook/index.md
+++ b/content/intro-to-storybook/index.md
@@ -64,6 +64,7 @@ contributors:
       detail: "Engineer at Squarespace",
     },
   ]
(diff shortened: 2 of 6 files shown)


content/intro-to-storybook/index.md
```diff
@@ -64,6 +64,7 @@ contributors:
       detail: "Engineer at Squarespace",
     },
   ]
+twitterShareText: "I’m learning Storybook! It’s a great dev tool for UI components."
```
kylesuss: This was the original messaging from before the refactor/updated design.

content/design-systems-for-developers/index.md
```diff
@@ -62,6 +62,7 @@ contributors: [
       detail: "UX Development at Shopify",
     },
 ]
+twitterShareText: "I’m learning about building design systems! They're great for scaling frontend code on large teams."
```
kylesuss: I just made this up. Happy to edit or just merge and have you edit, but I wanted to have this come through in a PR so you could see it @domyen

PR #145, opened 2019-10-02
Add loading indicator using Gatsby hooks for slow reqs

kylesuss:
Adds a loading indicator for page requests that are taking a while. Typically, this will only happen for users on slow networks.

https://www.loom.com/share/c37a41fa75644765a9a9e133cbe3b45a

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -1,7 +1,24 @@
-/**
- * Implement Gatsby's Browser APIs in this file.
- *
- * See: https://www.gatsbyjs.org/docs/browser-apis/
- */
+const React = require('react');
+const ReactDOM = require('react-dom');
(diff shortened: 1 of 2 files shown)


gatsby-browser.js
```diff
@@ -1,7 +1,24 @@
...
- * See: https://www.gatsbyjs.org/docs/browser-apis/
- */
+const React = require('react');
+const ReactDOM = require('react-dom');
+const { Spinner } = require('@storybook/design-system');
 
-// You can delete this file if you're not using it
+const LOADING_ID = '___loading';
+
+exports.onRouteUpdateDelayed = () => {
```
kylesuss: This hook only fires for slow transitions, not all the time. Gatsby says:

> Called when changing location is longer than 1 second.

PR #149, opened 2019-10-03
Typo in code snippet

beaucalvez:
Looks like there was a piece of HTML in that svn command


jonniebigodes:
wow, nice catch, looks like this one managed to slip through 👍

PR #151, opened 2019-10-04
Rename test variable

Tmarty:
Previously there was a disconnect between the code and variable name. This commit changes the variable name to be more in line with it's value and purpose.


tmeasday:
That makes sense, thanks @Tmarty

jonniebigodes:
I'll update the translation accordingly shortly

PR #152, opened 2019-10-04
Fix typo: GlobalStyle import path.

Beardles:
Fixing import path for `GlobalStyle` -- remove `/components/` from path.


domyen:
Thank you @Beardles

PR #159, opened 2019-10-07
We no longer need to import from storybook-chromatic!

tmeasday:


content/intro-to-storybook/angular/en/test.md
```diff
@@ -65,12 +65,11 @@ yarn add storybook-chromatic
...
 ```javascript
-import { configure } from "@storybook/angular";
-import "storybook-chromatic";
-import "../src/styles.less";
+import { configure } from '@storybook/angular';
+import '../src/styles.less';
 
 // automatically import all files ending in *.stories.ts
-const req = require.context("../src/", true, /.stories.ts$/);
+const req = require.context('../src/', true, /.stories.ts$/);
```
kylesuss: This is kinda an annoyance. I have manually turned off formatting on save for these MD files before when editing them, but I think this should be addressed eventually:

https://github.com/chromaui/learnstorybook.com/issues/162

domyen:
@tmeasday Do we need to remove the "Import chromatic into config" instruction as well?
image

tmeasday:
Yeah, fair point

PR #161, opened 2019-10-07
Fix width of GitHub nav item. Too small now that we have more stars

kylesuss:
Broken:

Screen Shot 2019-10-07 at 9 03 12 AM

Fixed in this PR:

Screen Shot 2019-10-07 at 9 03 08 AM

### DIFF
--- a/src/components/composite/Header.js
+++ b/src/components/composite/Header.js
@@ -121,7 +121,12 @@ const MobileMenuNavItem = styled(NavItem)`
 `;
 
 const GithubNavItem = styled(NavItem)`
-  width: 109px;
+  display: none;
+  min-width: 118px;
+
+  @media (min-width: 375px) {
+    display: inline-flex;
+  }
 `;
 
 const Nav = styled.div`


src/components/composite/Header.js
```diff
@@ -121,7 +121,12 @@ const MobileMenuNavItem = styled(NavItem)`
 `;
 
 const GithubNavItem = styled(NavItem)`
-  width: 109px;
+  display: none;
+  min-width: 118px;
```
kylesuss: Min width used this time in case it gets bigger. Ultimately, the width here is meant to decrease the "jumping" that happens when the GitHub badge initially loads since it doesn't take up space at first.

src/components/composite/Header.js
```diff
@@ -121,7 +121,12 @@ const MobileMenuNavItem = styled(NavItem)`
 `;
 
 const GithubNavItem = styled(NavItem)`
-  width: 109px;
+  display: none;
```
kylesuss: Hide it on really small devices. Otherwise, it sits on top of the logo.

PR #163, opened 2019-10-07
Remove extra semicolon in the global.js codeblock

Rekenna:
There was an extra semicolon in the pre-formatted code on line 83 which prevents that code from working; So I figured I would remove it.

Thanks for the write up guys! This is an excellent resource.


domyen:
Awesome thanks so much @Rekenna

kylesuss:
@Rekenna sorry I reverted this. I think you introduced a syntax error as the closing backtick was removed. Can you potentially create an issue with the error that you are experiencing that caused you to submit this PR? That would be appreciated. Thanks 🙇

PR #168, opened 2019-10-07
Upgrade design system to 1.1.2

kylesuss:
Upgrade the design system. I thought that some changes to the `Highlight` component might affect this repo w/ regard to font size, but it looks like we were explicit enough about that already in this repo. Therefore, no visual changes!


domyen:
*chefs kiss*

PR #169, opened 2019-10-07
Misplaced back tick

beaucalvez:
Removed misplaced back tick in the webpack.config.js snippet


domyen:
This looks good to me. @kylesuss can you sanity check?

PR #171, opened 2019-10-08
Upgrade design system, use GlobalStyle from design system

kylesuss:
Few things here:

1. Upgrade the design system to 1.1.6. Major thing this is doing is bringing in the `Highlight` component and global styling for code & pre elements that all the consumers will share.
2. Refactor some code here to use the `GlobalStyle` from the design system. Missed this somehow during the dev of the new layout.


domyen:
Dope looks good ty

Issue #162, opened 2019-10-07
Fix prettier settings for MD files or reformat all of them


kylesuss:
It seems that saving a MD file causes it to update for those that have prettier enabled. I think that is expected, but maybe we can do 1 big PR where all of the files are reformatted at once so it doesn't just happen sporadically in smaller PRs.

domyen:
Yea I agree

nickytonline:
First, thanks for all the great work. I can work on this. I can also add a precommit hook via Husky and lint-staged if you want. I did it over on the DEV repository a while back.

https://github.com/thepracticaldev/dev.to/blob/master/package.json#L18

domyen:
@nickytonline I'm into it! What do you think @kylesuss?

kylesuss:
Sounds great @nickytonline 💯 

We had a little convo going in a PR that might be of interest to you:

https://github.com/chromaui/learnstorybook.com/pull/175#issuecomment-540718986

We aren't sure that the prettier/eslint setup is correct with regard to markdown files. It feels odd that the markdown files are autoformatting to have double quotes whereas our other files are single quoted. There are other inconsistencies as well.

Would be amazing if you could look into that while you do this 💪 

As for husky, I am all for it. We have that over in the Storybook design system as well.

jonniebigodes:
Closing this down as the issue is no longer present.

PR #181, opened 2019-10-11
All markdown files are properly formatted now

nickytonline:
This PR implements #162, and adds a precommit hook via [lint-staged](https://github.com/okonet/lint-staged/) and [husky](https://github.com/typicode/husky). For the time being, I have only added an entry to lint-staged to format markdown (md) files. As well, all the md files have been formatted using the prettier settings that I copied from Storybook's settings.

Built everything locally and it looks good. I just did a smoke test navigating to some pages.

learnstorybook.com screeshot

Also, Skeletor approves of this PR.

Skeletor approves

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,5 @@
+const prettierConfig = require('./.prettierrc');
+
 module.exports = {
   parser: 'babel-eslint',
   extends: ['airbnb', 'prettier', 'plugin:jest/recommended'],
@@ -23,17 +25,7 @@ module.exports = {
         specialLink: ['to'],
       },
     ],
-    'prettier/prettier': [
-      'warn',
-      {
-        printWidth: 100,
-        tabWidth: 2,
-        bracketSpacing: true,
-        trailingComma: 'es5',
-        singleQuote: true,
-        jsxBracketSameLine: false,
-      },
-    ],
+    'prettier/prettier': ['warn', prettierConfig],
     'react/jsx-filename-extension': [
       1,
       {
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+yarn.lock
+package-lock.json
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "printWidth": 100,
+  "tabWidth": 2,
+  "bracketSpacing": true,
+  "trailingComma": "es5",
+  "singleQuote": true
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,0 @@
-{
-  "printWidth": 100,
-  "tabWidth": 2,
-  "bracketSpacing": true,
-  "trailingComma": "es5",
-  "singleQuote": true
-}
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,8 @@
+module.exports = {
+  printWidth: 100,
+  tabWidth: 2,
+  bracketSpacing: true,
+  trailingComma: 'es5',
+  singleQuote: true,
+  jsxBracketSameLine: false,
+};
--- a/content/design-systems-for-developers/index.md
+++ b/content/design-systems-for-developers/index.md
@@ -1,67 +1,68 @@
 ---
-title: "Design Systems for Developers"
-description: "Discover how to build and maintain design systems using Storybook."
-heroDescription: "A guide that teaches professional developers how to transform component libraries into design systems and set up the production infrastructure used by leading frontend teams."
+title: 'Design Systems for Developers'
+description: 'Discover how to build and maintain design systems using Storybook.'
+heroDescription: 'A guide that teaches professional developers how to transform component libraries into design systems and set up the production infrastructure used by leading frontend teams.'
 overview: "Design systems power the frontend teams of Shopify, IBM, Salesforce, Airbnb, Twitter,  and many more. This guide for professional developers examines how the smartest teams engineer design systems at scale and why they use the tools they use. We'll walk through setting up core services, libraries, and workflows to develop a design system from scratch."
 order: 2
-themeColor: "#0079FF"
-codeGithubUrl: "https://github.com/chromaui/learnstorybook-design-system"
+themeColor: '#0079FF'
+codeGithubUrl: 'https://github.com/chromaui/learnstorybook-design-system'
 heroAnimationName: null
 toc:
   [
-    "introduction",
-    "architecture",
-    "build",
-    "review",
-    "test",
-    "document",
-    "distribute",
-    "workflow",
-    "conclusion",
+    'introduction',
+    'architecture',
+    'build',
+    'review',
+    'test',
+    'document',
+    'distribute',
+    'workflow',
+    'conclusion',
   ]
-coverImagePath: "/guide-cover/design-system.svg"
-thumbImagePath: "/guide-thumb/design-system.svg"
-contributorCount: "+38"
+coverImagePath: '/guide-cover/design-system.svg'
+thumbImagePath: '/guide-thumb/design-system.svg'
+contributorCount: '+38'
 authors:
   [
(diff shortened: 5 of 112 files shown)


.prettierignore
```diff
@@ -0,0 +1,2 @@
+yarn.lock
```
nickytonline: We never want prettier to format these files.

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
nickytonline: Copied from https://github.com/storybookjs/storybook/blob/next/.prettierrc

package.json
```diff
@@ -40,6 +40,17 @@
     "storybook": "start-storybook -s ./static -p 6006",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
+  "husky": {
```
nickytonline: The precommit magic.

package.json
```diff
@@ -40,6 +40,17 @@
...
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.md": [
```
nickytonline: Other file types can be added here in another PR.

.prettierrc
kylesuss: Is it possible to make this formatting work with the prettier config that we have already here?

https://github.com/chromaui/chromatic/blob/next/.eslintrc.js#L31

The eslint config is already setup with these params.

nickytonline: Oops, didn’t see that. I will update it later tonight.

.eslintrc.js
```diff
@@ -23,17 +25,7 @@ module.exports = {
...
-      {
-        printWidth: 100,
-        tabWidth: 2,
-        bracketSpacing: true,
-        trailingComma: 'es5',
-        singleQuote: true,
-        jsxBracketSameLine: false,
-      },
-    ],
+    'prettier/prettier': ['warn', prettierConfig],
```
nickytonline: This is imported now, so that files that are non-JS can use the same prettier config.

.prettierrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
nickytonline: This is the config that get's imported into .eslintrc.js

nickytonline:
Alright, I updated things to use the existing prettier config. Let me know if that's all good. That was the first time I've seen a prettier config in an eslint config which is why I didn't think to look there. ;)

.eslintrc.js
nickytonline: I am curious as to why eslint is configured to only warn about prettier. I imagine once a config for JS files is added to the lint-staged configuration, this would no longer be required?

nickytonline:
Yeah, sharing seemed to make the most sense to me for the prettier confit.

I’ll rerun the batch prettier formatting for the non-prettier formatted *.md files. I think I used the wrong glob when I ran it, so it didn’t format all the files. I’ll update the PR this afternoon.

And as you mentioned, the only way for markdown files to get formatted is on the precommit hook. If you have an editor like VS Code you can have them formatted on save with the prettier extension, but there is no guarantee that everyone contributing would be using VS Code, so the precommit hook is a fail safe.

As mentioned previously, once this PR is merged, we can use the precommit hook for other file types as well: *.jsx, *.js, *.json etc.

nickytonline:
So the reason that particular file would not prettify is because of the location of the array square brackets for contributors. I lined them up with those of the `authors` section and then prettier ran as usual. That was not obvious. It looks like that is the only file that wasn't prettified. I'll push that change momentarily.

nickytonline:
Hmm even with that fix there is still parts that did not convert to single quotes. I'll continue to investigate.

nickytonline:
Ahh, I see what it is. It's mentioned in a prettier issue. See https://github.com/prettier/prettier/tree/21733e441dc01f7d85b483edee92b7e7507bfd9a#quotes. It's by design.

> If the number of quotes outweighs the other quote, the quote which is less used will be used to format the string - Example: "I'm double quoted" results in "I'm double quoted" and "This \\"example\\" is single quoted" results in 'This "example" is single quoted'.

nickytonline:
I went through the other markdown files that have double quotes, and they are either respecting the rule I mentioned previously or they are markup and stay as double quotes instead of single quotes.

content/design-systems-for-developers/index.md
```diff
@@ -1,67 +1,68 @@
 ---
-title: "Design Systems for Developers"
-description: "Discover how to build and maintain design systems using Storybook."
-heroDescription: "A guide that teaches professional developers how to transform component libraries into design systems and set up the production infrastructure used by leading frontend teams."
+title: 'Design Systems for Developers'
+description: 'Discover how to build and maintain design systems using Storybook.'
+heroDescription: 'A guide that teaches professional developers how to transform component libraries into design systems and set up the production infrastructure used by leading frontend teams.'
 overview: "Design systems power the frontend teams of Shopify, IBM, Salesforce, Airbnb, Twitter,  and many more. This guide for professional developers examines how the smartest teams engineer design systems at scale and why they use the tools they use. We'll walk through setting up core services, libraries, and workflows to develop a design system from scratch."
```
nickytonline: As mentioned in the main convo, this is how prettier works. Since there is a single quote in. the text, it will keep the double quotes for the string.

.eslintrc.js
kylesuss: I think the initial logic was that we didn't want to discourage people from contributing because of it. But yeah, given the changes you have implemented here which help reinforce the standards, I think we can change it eventually.

.prettierignore
```diff
@@ -0,0 +1,2 @@
...
+package-lock.json
```
kylesuss: Nice. Haha we aren't supposed to have both of these lockfiles, so I will submit a small PR to remove the `package-lock.json`.

content/intro-to-storybook/react/zh-TW/composite-component.md
```diff
@@ -1,19 +1,19 @@
...
-tocTitle: "合成元件"
-description: "使用更簡單的元件 組裝複合元件"
+title: '組裝複合元件'
+tocTitle: '合成元件'
+description: '使用更簡單的元件 組裝複合元件'
 commit: '8db511e'
 ---
 
-上一章我們構建了第一個元件; 本章 我們學習 擴充套件構建TaskList的任務列表. 讓我們將 元件組合 在一起,看看在引入更多複雜性時會發生什麼.
+上一章我們構建了第一個元件; 本章 我們學習 擴充套件構建 TaskList 的任務列表. 讓我們將 元件組合 在一起,看看在引入更多複雜性時會發生什麼.
```
kylesuss: @danielduan or @chinanf-boy -- pinging you here since you contributed the original Chineses translations. Prettier is formatting some of these files slightly differently. I recognize there are 2 Chinese translations and it appears to happen in both of them:

Seems like the common change is to insert a space around English words and Chinese characters. So:

`擴充套件構建TaskList的任務列表`

becomes

`擴充套件構建 TaskList 的任務列表`

The same happens in the `zh-CN` translation. Does this "break" the translation at all? Or is it an acceptable change? If it breaks things, I think we can probably ignore these files in Prettier.

Thanks 🙇

nickytonline:
Sounds good. I’ll get to this later this evening.

nickytonline:
Alright, I've rebased and fixed the merge conflicts @kylesuss. So this should be good to go. As you mentioned, just waiting to hear about the Chinese translations.

domyen:
@kylesuss @nickytonline I suggest we move forward with auto-formatting Chinese. According to this [article](https://www.hillslearning.com/2011/06/15/writing-in-chinese-spacing/) it's fine to add a space between English words and Chinese characters.

kylesuss:
@domyen nice find. Interestingly, that article seems to imply that the old way was wrong! Ha. Yes let's move forward on this and circle back if need be.

PR #190, opened 2019-10-16
Update example code to be equal like the English chapter

dulcehc:
I just updated the examples to be equal like in the English guide.


domyen:
Thank you @dulcehc

PR #195, opened 2019-10-18
Upgrade design system to 1.1.7, update yml -> yaml for Highlight

kylesuss:
Upgrade the design system. The major addition is yaml syntax highlighting  which was added here: https://github.com/storybookjs/design-system/pull/97

Here it is in action on one of the guides:
Screen Shot 2019-10-18 at 1 25 07 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.10.0
+      - image: circleci/node:8.12.0
 
     working_directory: ~/repo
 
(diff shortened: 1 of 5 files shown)


content/design-systems-for-developers/react/en/test.md
```diff
@@ -89,7 +89,7 @@ Visual testing helps identify UI changes in Storybook. Review the changes to con
 
 Let’s add visual testing to the continuous integration job. Open `.circleci/config.yml` and add the test command.
 
-```yml
+```yaml
```
kylesuss: I think there is something about how we are using the library that it required the full name. We have to tell prism js which languages we want to use so we don't have to load them all and I manually said `yaml` so maybe the library requires that you reference it in that way from there on out.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.10.0
+      - image: circleci/node:8.12.0
```
kylesuss: One of the nested deps required this upgrade.

domyen:
Nice

Issue #192, opened 2019-10-16
Link languages on GuideScreen to respective guides


domyen:
Link the languages listed on guide screen to their respective guides. This makes the languages more discoverable and saves non-English speakers from extra clicks. 

<img width="730" alt="Screen Shot 2019-10-16 at 10 19 33 AM" src="https://user-images.githubusercontent.com/263385/66927877-91e5d100-effe-11e9-847b-83a069b62185.png">

jonniebigodes:
@domyen i'm going to make a pull request soon addressing this issue. Also one thing before i proceed further, do you want the component story files updated as well?

domyen:
Yea! Since we do Component-Driven Development the [Hero component's](https://www.chromaticqa.com/component?appId=5cf841a3f3e3d200208ffc74&name=Screens%7CGuideScreen%2FHero&buildNumber=198&specName=default) story should get updated to reflect the new behavior.

jonniebigodes:
@domyen i'll do the final round of checking tomorrow as it's almost 2 am where i'm at. And create the pr detailing what i've done. Sounds good?

jonniebigodes:
Fixed with #204

PR #204, opened 2019-10-24
Fix for issue #192

jonniebigodes:
This PR addresses issue #192 . 

Below is a list of the changes made to fix the issue:

- In `src/components/screens/GuideScreen/index.js`:
  - Moved `getLanguageName` function call from the `Hero` component to this component.
  - Modified the `languages` array from a array of strings, to be a array of objects that contain the language and the translation path. The  name is obtained from calling the `getlanguageName` function and the path of the translation is constructed by retrieving the `slug` from gatsby's special prop `pageContext`. As per present implementation it will default to the react tutorial.
  - Updated the PropTypes for `translationPages`, to prevent warnings and reflect the current graphql result.
- In `src/components/screens/GuideScreen/index.stories.js`:
  - Added `pageContext` to the props object, per implementation above. So that the stories don't break.
- In `src/components/screens/GuideScreen/Hero.js`:
  - Removed `getLanguageName ` function call as it makes more sense to use it in the `src/components/screens/GuideScreen/index.js` component. making this component even more simple.
  - Styled the links.
- `src/components/screens/GuideScreen/Hero.stories.js`
   - A array containing possible languages and tutorial translations was created. And some extra stories aswell to check how the component would react and display the information. 

Feel free to provide feedback

### DIFF
--- a/src/components/screens/GuideScreen/Hero.js
+++ b/src/components/screens/GuideScreen/Hero.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
 import { Button, styles } from '@storybook/design-system';
-import { rgba } from 'polished';
 import GatsbyLink from '../../basics/GatsbyLink';
 import Stat from '../../basics/Stat';
 import * as animations from '../../../styles/animations';
@@ -107,17 +106,6 @@ const LanguagesLabel = styled.span`
 `;
 const LanguageLinkStyles = `
   && {
-    color: ${styles.color.lightest};
-    &:hover {
-      color: ${rgba(styles.color.lightest, 0.5)};
-    }
-    &:visited{
-      color: ${styles.color.lightest};
-    }
-    &:active {
-      color: ${rgba(styles.color.lightest, 1)};
-    }
-   
     &:not(:last-child):after{
       content:', ';
       white-space: pre;
@@ -202,7 +190,7 @@ const Hero = ({
           <Languages>
             <LanguagesLabel>Languages: </LanguagesLabel>
             {languages.map(language => (
-              <LanguagesLink key={`lang_link_${language.name}`} to={language.tutorial}>
+              <LanguagesLink inverse key={`lang_link_${language.name}`} to={language.tutorial}>
                 {language.name}
               </LanguagesLink>
             ))}
--- a/src/components/screens/GuideScreen/Hero.js
+++ b/src/components/screens/GuideScreen/Hero.js
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
 import { Button, styles } from '@storybook/design-system';
+import { rgba } from 'polished';
 import GatsbyLink from '../../basics/GatsbyLink';
 import Stat from '../../basics/Stat';
-import getLanguageName from '../../../lib/getLanguageName';
 import * as animations from '../../../styles/animations';
 
 const { breakpoint, color, pageMargins, spacing, typography } = styles;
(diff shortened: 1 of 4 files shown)


src/components/screens/GuideScreen/Hero.js
```diff
@@ -105,6 +105,29 @@ const Languages = styled.div`
...
+    color: ${styles.color.lightest};
+    &:hover {
+      color: ${rgba(styles.color.lightest, 0.5)};
+    }
+    &:visited{
+      color: ${styles.color.lightest};
+    }
+    &:active {
+      color: ${rgba(styles.color.lightest, 1)};
+    }
```
kylesuss: I think we should take these custom link styles out and instead just pass the `inverse` prop to `GatsbyLink`:

```javascript
<LanguagesLink inverse
...
```

They are similar enough that it feels like using the design system's prebuilt styles is the right approach. Thoughts?

src/components/screens/GuideScreen/Hero.js
```diff
@@ -207,7 +230,12 @@ Hero.propTypes = {
   description: PropTypes.string,
   heroAnimationName: PropTypes.string,
   imagePath: PropTypes.string,
-  languages: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
+  languages: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string,
+      tutorial: PropTypes.string,
+    })
```
kylesuss: Is this required? My hunch is that it is.

```suggestion
    }).isRequired,
```

src/components/screens/GuideScreen/index.js
```diff
@@ -202,8 +209,22 @@ Guide.propTypes = {
...
+              language: PropTypes.string,
+              slug: PropTypes.string,
+            }),
+          }),
+        })
+      ),
+    }),
   }).isRequired,
+  pageContext: PropTypes.shape({
+    slug: PropTypes.string,
```
kylesuss: Nice! You know I didn't even realize Gatsby offered this data. Good to know! Since it will always be around, can we require it?

```suggestion
    slug: PropTypes.string.isRequired,
```

src/components/screens/GuideScreen/index.js
```diff
@@ -202,8 +209,22 @@ Guide.propTypes = {
...
+  }),
```
kylesuss:
```suggestion
  }).isRequired,
```

kylesuss:
@jonniebigodes left a few minor comments, though this looks great! Let's discuss the comments and then we can get this merged once that happens.

src/components/screens/GuideScreen/index.js
jonniebigodes: @kylesuss yes it does, when the `createPages` action is running in `gatsby-node.js` whatever is injected via gatsby's special prop `context` can be accessed via `pageContext` in the page, in a nutshell that's how the page queries extract the parameters. I'm going to update the component to reflect the change.

src/components/screens/GuideScreen/Hero.js
jonniebigodes: you're right it should.

src/components/screens/GuideScreen/Hero.js
jonniebigodes: Sounds good to me! I wondered about if this could be better handled with the design system. I took a peek at the code in the design system and wasn't able to find a reference to it, so i implemented the "custom styling", i have t take a deeper look at the code it seems. I'm updating the component accordingly

jonniebigodes:
@kylesuss thank you for the feedback, The changes were made per your feedback. Let me know if anything else is required.

src/components/screens/GuideScreen/Hero.js
jonniebigodes: i'll update accordingly and make the prop required, as it should, this will always be required, sorry about that, it was a oversight from my part.

src/components/screens/GuideScreen/index.js
jonniebigodes: Once again i missed that one

src/components/screens/GuideScreen/Hero.js
kylesuss: No worries! I figured that was the case. For reference, we have the deployed Storybook for the design system here:

https://storybook-design-system.netlify.com/?path=/docs/*

It is also linked in the [repo](https://github.com/storybookjs/design-system).

src/components/screens/GuideScreen/Hero.js
kylesuss: No worries at all.

jonniebigodes:
@kylesuss no need to thank, glad i was able to help. @domyen I'm going to close the issue associated with the pr if you don't mind.

src/components/screens/GuideScreen/Hero.js
jonniebigodes: gonna bookmark it for future reference. Thanks for the links.

domyen:
Closing the issue sounds good!

PR #207, opened 2019-10-28
Update creating-addons.md

amalv:
Use content component instead of placeholder for the example creating the content component.


jonniebigodes:
@amalv nice catch. Thank for you for the small but very important contribution to the document.

PR #209, opened 2019-11-03
Fix some punctuation typos

tibbes:


jonniebigodes:
@tibbes thanks for this small but very important contribution to the documentation. Looks like this one slipped though.

PR #210, opened 2019-11-04
Update data.md to add redux dependency

asgabe:
A minor fix when adding `react-redux` dependency without `redux` library


jonniebigodes:
@domyen @tmeasday @kylesuss when you can merge this, the change is correct, did a quick search and awhile ago the spanish translation was updated to reflect this and as the original document. 

See #190 for the exact same change.

@asgabe nice catch on your end. Thank you very much for the contribution, a very important one 👍

PR #220, opened 2019-11-16
update .storybook/config.js instructions

rphutchinson:
users are instructed to update .storybook/config.js to reference nonexistent `src/components` directory. Updating documentation to match actual source tree `src`


jonniebigodes:
Nice catch, you're right, the code example points the correct way and the text to a wrong way. The `src/components` path is commonly used in intro to storybook tutorial. Thank you for catching this small but very important typo. 👍

PR #232, opened 2019-12-15
Import actions in stories and correct reference to CSS in config

patrickgrey:
I'm running on Windows 10. On "npm run storybook" I get "action not defined" feedback in browser. Importing actions fixes this. Previously we add style references to global.css but in the config index.css is referenced. I changed the reference to global.css.

### DIFF
--- a/content/intro-to-storybook/svelte/en/simple-component.md
+++ b/content/intro-to-storybook/svelte/en/simple-component.md
@@ -66,6 +66,7 @@ Below we build out Task’s three test states in the story file:
 
 ```javascript
 // src/components/Task.stories.js
+import { action } from "@storybook/addon-actions";
 import Task from './Task.svelte';
 
 export default {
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@patrickgrey you're right..when the pr was going over a revision process some changes were required and with that i accidently removed that line. Nice catch on your end. Adding it should fix it, sorry for that.  And the css change is not required. The css is going to be picked up either way as you can see in the code base used [here](https://github.com/jonniebigodes/intro-storybook-svelte) and a live version [here](https://jonniebigodes-storybook-svelte.netlify.com/?path=/story/pureinboxscreen--standard). I think we should leave it as is. Thank you for catching this and once again sorry for the mishap

jonniebigodes:
@tmeasday or @domyen or @kylesuss this can be closed as it's now fixed with the recent change into the documentation

tmeasday:
Thanks @jonniebigodes

PR #234, opened 2019-12-22
Moving `component` property to default export

davidhorm:
Moving `component` property to align with documentation at https://storybook.js.org/docs/formats/component-story-format/#default-export

FYI - I just noticed this discrepancy when I was looking at https://github.com/storybookjs/storybook/blob/master/addons/docs/docs/docspage.md#component-parameter. But as far as I can tell having `parameters: { component: ... } }` was able to extract the JSDoc still.

Also, I noticed `component` being set differently in the Svelte examples. But I left that documentation alone since I don't know anything about Svelte.

### DIFF
--- a/content/design-systems-for-developers/react/en/document.md
+++ b/content/design-systems-for-developers/react/en/document.md
@@ -62,9 +62,9 @@ Start by adding more metadata that explains what the component does. In `src/Ava
 ```javascript
 export default {
   title: 'Design System|Avatar',
+  component: Avatar,
 
   parameters: {
-    component: Avatar,
     componentSubtitle: 'Displays an image that represents a user or organization',
   },
 };


jonniebigodes:
@davidhorm i'll defer the change you're introducing to either @domyen or @tmeasday as they are the content creators.

Regarding this item:

> Also, I noticed component being set differently in the Svelte examples. But I left that documentation alone since I don't know anything about Svelte.

I wrote the tutorial and @domyen and @tmeasday have reviewed it. The documentation is accurate. That's the way you use both tools together. Despite the drive for unification on how the stories are implemented in Storybook with csf, adding support for various frameworks has it's quirks. With that for instance vue will be setup a bit differently and the same goes for angular.

tmeasday:
I think this PR is correct and this is a mistake. Can you confirm @shilman?

jonniebigodes:
@davidhorm truly sorry for not giving further feedback on this. But with the latest updates in the Design Systems tutorial this pull request was addressed already. Hope you don't mind, but i'm going to close this pull request. 
 

Stay safe.

PR #235, opened 2019-12-23
update code snippet for initial config.js

jacknevitt:
`requireContext` does not get introduced until later


jonniebigodes:
@jacknevitt you're correct. And sorry for introducing this oversight, as of late the documentation is being rewritten to incorporate the new way of writting stories. And it seems that this one slipped through. @domyen or @kylesuss when you're able merge this if you don't mind.

jacknevitt:
Thanks for getting this fixed so quick

PR #239, opened 2020-01-03
Update get-started.md

jadestern:
Fixed command for omitted option.


jonniebigodes:
@jadestern thank you very much for this small but rather important fix. This one slipped through during the review process. 👍

PR #242, opened 2020-01-07
Maybe typo

DaleSeo:
It looks like a typo here.

### DIFF
--- a/content/intro-to-storybook/react/en/composite-component.md
+++ b/content/intro-to-storybook/react/en/composite-component.md
@@ -9,7 +9,7 @@ Last chapter we built our first component; this chapter extends what we learned 
 
 ## Tasklist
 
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@DaleSeo i think not a typo. If you check the image below that phrase and read it up again you'll see that the intention is for you (the reader) that the TaskList will yeld 2 scenarios, one scenarion where there's only default items, not pinned, or archived. And the second scenario where there both the default items and pinned items coexist.

DaleSeo:
@jonniebigodes Oops, sorry. I think that makes sense. Will close this PR then.

jonniebigodes:
@DaleSeo no problem whatsoever, i encountered the same issue you had. After looking at the image is when it clicked. I was reading phrase and i was about to create a pr to address this and then i looked a bit lower and saw the image. But it goes without saying, thank you for the contribution

PR #244, opened 2020-01-07
actionsData -> actionData

DaleSeo:
`actionData` is exported in `Task.stories.js` so `actionData` should be imported instead of `actionsData`.


jonniebigodes:
@DaleSeo i've checked the documentation and it seems correct. In the simple component section the `actionsData` is consumed by the story for the Task and in the composite component section extends it and uses it as well.

DaleSeo:
Oops, sorry. Did you guys change recently? Maybe I was confused. Sorry about that wrong call again. Will close this PR.

jonniebigodes:
@DaleSeo no problem. The documentation and code base is being updated incrementally and with that probably some fixes were applied.

PR #251, opened 2020-01-14
feat: update documentation to match version 5.3

jonniebigodes:
With this pr the svelte tutorial is updated to version 5.3 of Storybook.

What was done:
- The Svelte code examples were updated as prettier is introducing some changes that mislead readers and lead to errors.
- The CSS change that was mentioned in #232 is addressed and with that that pr can be closed as technically the configuration is now obsolete.
- Applied some minor tweaks to the wording used so that it aligns itself with the new configuration introduced in 5.3.

I'm adding the `DO NOT MERGE` label for now, as the create addons section is not working with 5.3 and i would like some input from your end. I have a [open issue](https://github.com/storybookjs/storybook/issues/9421)  in the storybook repo addressing it. But until that is resolved i would like to know on how to proceed?
Will we focus on pushing the documentation and with that cut the create addons section and introducing a navigation issue with the site or we wait for it to be concluded.

@tmeasday and @domyen when you're able check the Simple component's wording, more specifically the config section so that we can align the documentation. I went with a simpler wording.

Feel free to provide feedback


content/intro-to-storybook/svelte/en/simple-component.md
```diff
@@ -148,16 +152,27 @@ When creating a story we use a base task (`task`) to build out the shape of the
...
+Start by changing your Storybook configuration file (`.storybook/main.js`) to the following:
 
 ```javascript
-// .storybook/config.js
-import { configure } from '@storybook/svelte';
-import '../src/index.css';
-configure(require.context('../src/components', true, /\.stories\.js$/), module);
+// .storybook/main.js
+module.exports = {
+  stories: ['../src/components/**/*.stories.js'],
```
tmeasday: Should we do this, or just remove the original `src/stories` files?

jonniebigodes: Remove the original stories from the Storybook package? Is that what you mean? If that's the case i think it's probably not a good thing, not only for the user/reader, but also in terms of documentation as it means to make some changes to screenshots and so on. I think leaving a bare bones example in the package is a nice touch to the end user and letting them know with  this small change is a safer route. But i'll leave it to you and the rest of the team on how to proceed

jonniebigodes:
@tmeasday there's still time to make the change 😃

content/intro-to-storybook/svelte/en/simple-component.md
tmeasday: Oh, sorry, what I meant was referencing the comment you made on my PR: https://github.com/chromaui/learnstorybook.com/pull/249#discussion_r366488733

If we set `stories: ['../src/**/*.stories.js']`, it will pick up the stories put down by the `sb init` command, right? So we could do that, but just delete those files at this point too.

tmeasday: Oh, sorry, my bad, I see that SB sets it to `'../stories/**/*.stories.js', so we'd need to change it in either case.

So the question is: should we change it to `../src/**/*.stories.js' or `../src/components/**/*.stories.js'?.

jonniebigodes: I think we should go with `../src/components/**/*.stories.js` as the majority of the documentation leads the user/reader into that pattern, all of the components used are placed inside the `components` folder. I don't know if it's significant to Storybook itself, but having it just instrospect that folder will tend to speed up the boot process. Don't you think?

jonniebigodes:
With the latest commit the svelte and other documentation that uses npm as a package manager will have the correct image already present. I grabbed it from the angular documentation that i'm working on and added it here and updated the filename. As this pr is way more advanced than the angular one at least the image and documentation will be correct and revert back a some misleading introduced by a recent pr

content/intro-to-storybook/svelte/en/simple-component.md
tmeasday: Sure, ok, let's move forward with that, I'm sure users can figure it out if they want to put stories elsewhere in their own projects.

tmeasday:
@jonniebigodes I think we are good to merge this then? Shall I update the other PR with similar changes to this?

jonniebigodes:
@tmeasday sure go for it. I'll update the create addons section incrementally as soon as the pr for the fix with the addons comes into play. Also yes, go for it, use this as a baseline and start updating it. As you have already changed the get started and the simple component sections apply the changes, i'll update the rest of the documentation incrementally. One thing if you don't mind. As i have a pr opened up for the angular documentation and if you don't mind, revert back the changes you've made for that version. I'll update it based on this one and we move from there. Sounds good?

tmeasday:
I'll update my PR tomorrow, hopefully after this one is merged.

jonniebigodes:
I've just bumped Storybook to 5.3.4 and still have the same issue

jonniebigodes:
@tmeasday can you wait until i submit another commit. I was reading this once again and i need to update some items, as they are inducing a error.

PR #255, opened 2020-01-16
prop-types is unnecessary

DaleSeo:
Since `prop-types` is already included in `create-react-app`, I guess it doesn't need to be installed explicitly again?

### DIFF
--- a/content/design-systems-for-developers/react/en/architecture.md
+++ b/content/design-systems-for-developers/react/en/architecture.md
@@ -101,7 +101,7 @@ For the code sample, we’ve selected a subset of these components to make reaso
 We’ll also need to update dependencies that our components rely on.
 
 ```bash
-yarn add prop-types styled-components polished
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@DaleSeo i was thinking about it and i see your point, but as this ecosystem is know for it's volatility and with that cra could implement some change and drop the package altogether leaving it in could work as plus, also if you install it like it's documented it will be pulled in from the cache. But i'll defer to @domyen on how to proceed as he's co-author of the documentation.

domyen:
I'm fine with the explicit install of `prop-types` even though it's included in CRA. I think it feels less "magical" which is more transparent for folks that are following along.

Agree with @jonniebigodes that we can/should be defensive about upstream changes.

DaleSeo:
@jonniebigodes @domyen Thanks for the careful consideration!

PR #228, opened 2019-12-07
Chore: update angular documentation

jonniebigodes:
With this pr, the intro to storybook tutorial for angular is updated.  It was long overdue, and for that i would like to apologize to @geromegrignon for putting this on the "back burner" (pardon the bad pun).

It will close #213 #104  #227 and #60 .

The updated code is in [here](https://github.com/jonniebigodes/intro-storybook-5_3-angular)

Some caveats:
- I've added the do not merge label for now as there's still a lingering issue with angular and storybook as mentioned in [here](https://github.com/storybookjs/storybook/issues/8853) i'm going to test it out shortly, but for now i would like @geromegrignon 's input on this, as he expressed some knowledge in angular, probably more even than me. 
- In the deploy section, the settings image for netlify was updated, to avoid confusion, preventing the build from completing. as it's mentioning yarn and the tutorial is "operating" with npm and netlify will pick up on `package-lock.json` file and go from there.
- Added a message in the translations available, so that when the reader picks up on it. It will know that they are not updated yet. Probably bring in more and more people to help out with the translations.
- The code examples for stories are not using CSF. I think that could be addressed a bit later in a subsequent pr.

PR #257, opened 2020-01-19
Feat update vue pt translation

jonniebigodes:
With this pr the vue portuguese translation was updated to reflect the changes introduced with the release of Storybook >=5.3.

Also i've updated the deployment section's image to fix the misleading that was recently introduced in a pr. The npm version is already fixed, all that was missing was this one. 

Also the composite component's code snippet was fixed as it was wrong and for that i'm really sorry. As soon as the @tmeasday merges his pr i'll adjust accordingly the original version.

@domyen as per requested, so far the extension is good, asides from a couple of thing that need to be addressed, which i've mentioned in [here](https://github.com/oslabs-beta/aesop-for-storybook/issues/9). In my personal opinion this resource could be a great addition to the documentation when those issues are fixed. I'll let you and @tmeasday decide amongst yourselves on how to proceed and we move from there. Sounds good?

Feel free to provide feedback


jonniebigodes:
i think the angular tutorial was pulled in due to the fact that i "fumble fingered"(pardon the bad pun) from the angular translation branch i have locally instead of creating a branch from the master branch.... the angular updates is #228, To avoid "murking the waters" if you agree we close this and delete the branch and i'll start from a clean slate, doing things properly? Sounds good

jonniebigodes:
i'm going to close this pr as it seems that i've "fumbled fingered"(pardon the bad pun) when creating the branch for it. Instead of starting from the master, the branch was created from the branch were the angular documentation was being updated on. And to avoid "murking the waters" it would be best if this is done from a clean slate. I'll make a follow up pr afterwards.

PR #250, opened 2020-01-14
Update design-system-for-developers/react/en with triconfig

tmeasday:
(Base is currently set to `221-use-mainjs`)

Updated design systems to use `main.js` et al. Haven't actually run the code.

PR #258, opened 2020-01-20
Delete unused import

DaleSeo:
This line seems to be irrelevant.


jonniebigodes:
@DaleSeo you're right, as of some recent change in Storybook, adding the import is no longer necessary, It will be picked up by Storybook. But i think that it would be beneficial to the documentation  to evolve if this is adjusted to the current version. As the snippet is technically not  accurate anymore. And it's being updated in #250. But i'll defer to both @tmeasday and @domyen once again as they are the authors of the said documentation

tmeasday:
Thanks!

PR #259, opened 2020-01-20
Merge pull request #259 from chromaui/fix-translation-navigation

kylesuss:
The nav logic on the chapter screen assumed that all of the translated guides would have all chapters accounted for. This PR fixes that assumption so that we can correctly pick the "next" page to link to.


domyen:
Thanks @kylesuss and @jonniebigodes

PR #261, opened 2020-01-20
Use '@storybook/addon-docs/preset' instead of '@storybook/addon-docs/…

DaleSeo:
Just encountered the following warning that can be fixed by this PR.

```
DeprecationWarning: Framework-specific presets are no longer-needed as of Storybook 5.3 and will be removed in 6.0.

Please use '@storybook/addon-docs/preset' instead of '@storybook/addon-docs/react/preset'.
```


tmeasday:
Thanks @DaleSeo

PR #262, opened 2020-01-21
Move "Sharing Addons" to creating-addons.md, p.2/2

jonathanherdt:
The part "Sharing Addons" mentions that deploying will be handled in the next chapter. This is wrong as it is handled after the chapter `creating-addons`.


domyen:
@jonathanherdt thanks for taking the initiative here! I reworded both chapter's conclusions to flow better. What do you think?

jonathanherdt:
Looks awesome! 😍

PR #264, opened 2020-01-21
Add missing bracket in code

alonabas:
There is missing bracket in this function - if one copies it like this it throws exception.

### DIFF
--- a/content/intro-to-storybook/react/en/creating-addons.md
+++ b/content/intro-to-storybook/react/en/creating-addons.md
@@ -235,7 +235,7 @@ import { styled } from "@storybook/theming";
 
 const getUrl = input => {
   return typeof input === "string" ? input : input.url;
-
+}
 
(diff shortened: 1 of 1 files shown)


content/intro-to-storybook/react/en/creating-addons.md
```diff
@@ -235,7 +235,7 @@ import { styled } from "@storybook/theming";
 
 const getUrl = input => {
   return typeof input === "string" ? input : input.url;
-
+}
```
domyen: Does this need a `;`?

domyen:
Thanks for the PR @alonabas 🙏. Had one comment on syntax.

content/intro-to-storybook/react/en/creating-addons.md
alonabas: Right, added.

PR #266, opened 2020-01-22
1. Update task prop default value. 2. Change PureTaskList on TaskList.

sshiling:
1. The default value for task prop in Stories must be returned from the factory function (that's why it fails on the Jest tests).
2. PureTaskList is from the next lesson. Here it should be TaskList.

### DIFF
--- a/content/intro-to-storybook/vue/en/composite-component.md
+++ b/content/intro-to-storybook/vue/en/composite-component.md
@@ -101,36 +101,36 @@ export const withPinnedTasksData = [
 
 // default TaskList state
 export const Default = () => ({
-  components: { PureTaskList },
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@sshiling nice catch in those, regarding the first one, sorry but it seems i've copied over the wrong code snippet. For the second one the tests don't fail, The snapshots go through, vue cli throws a warning mentioning the default factory. I'm going to check this against my code base for the tutorial and see if it fixes the warning. And if does excelent

sshiling:
Yes, you're right. It's not the snapshot test error. It's the console error (Vue warn). The factory function will fix this warning.

jonniebigodes:
@domyen and @tmeasday if you both don't mind, i'm going to close this in favor of #271.

PR #267, opened 2020-01-22
Fix tutorial source code elements.

sshiling:
1. As described in the title.
2. task prop default value must be returned from the factory function (that's why it fails on the Jest tests).

### DIFF
--- a/content/intro-to-storybook/vue/en/simple-component.md
+++ b/content/intro-to-storybook/vue/en/simple-component.md
@@ -29,7 +29,7 @@ We’ll begin with the baseline implementation of the `Task`, simply taking in t
 ```html
 <template>
   <div class="list-item">
-    <input type="text" :readonly="true" :value="this.task.title" />
+    <input type="text" :readonly="true" :value="task.title" />
   </div>
 </template>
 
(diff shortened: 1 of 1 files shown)


content/intro-to-storybook/vue/en/simple-component.md
```diff
@@ -79,7 +79,7 @@ export const Default = () => ({
   template: taskTemplate,
   props: {
     task: {
-      default: taskData,
+      default: () => (taskData),
```
jonniebigodes: @sshiling we can do without the extra parenthesis, to avoid having prettier adding unnecessary warnings. I've just checked and all three modes work without it.

sshiling: @jonniebigodes okay. Thank you for your detailed explanations.

jonniebigodes:
@sshiling i was going about this and your other pr #266 and  if you could consolidate the changes and feedback that exists in these two into a single pr  and with that update the rest of the code snippets not only for this section, but also for the following sections:
- Composite component
- Data
- Using addons

I think it would be a invaluable addition to the documentation. What do you think? Sounds good?

sshiling:
@jonniebigodes it's great. Thanks. I'll do it in a few days.

jonniebigodes:
@sshiling no need to thank whatsoever. Looking forward for it

jonniebigodes:
@domyen and @tmeasday if you both don't mind, i'm going to close this in favor of #271.

PR #268, opened 2020-01-23
Correction in the translation of the term `utilizador` to` usuário` for pt language

angelolucas:
Correction in the translation of the term `utilizador` to`usuário` for pt language.
This change makes sense for Brazilian and Portugal portuguese.

### DIFF
--- a/content/design-systems-for-developers/react/pt/architecture.md
+++ b/content/design-systems-for-developers/react/pt/architecture.md
@@ -13,7 +13,7 @@ Em grandes empresas, este exercício é feito em conjunção com as equipas de d
 
 Se trabalha numa equipa de desenvolvimento, já deve ter reparado que equipas grandes não são muito eficientes. Falhas de comunicação aumentam à medida que as equipas crescem. Os padrões do IU existentes ficam por documentar ou perdem-se por completo. O que significa que os programadores passam mais tempo a reinventar a roda ao invés de construir novas funcionalidades. Com o decorrer do tempo os projetos estão cheios de componentes que são usados uma única vez.
 
-Nós demos de caras com este problema. Apesar das melhores intenções de uma equipa experiente, os componentes de interface de utilizador eram constantemente reconstruídos, ou colados. Os padrões de IU que deveriam ser os mesmos, começaram a divergir em aparência e funcionalidade. Cada componente tornou-se um floco de neve único, o que fez com que se tornasse impossível para os novos programadores distinguirem qual a fonte de verdade, quanto mais contribuir.
+Nós demos de caras com este problema. Apesar das melhores intenções de uma equipa experiente, os componentes de interface de usuário eram constantemente reconstruídos, ou colados. Os padrões de IU que deveriam ser os mesmos, começaram a divergir em aparência e funcionalidade. Cada componente tornou-se um floco de neve único, o que fez com que se tornasse impossível para os novos programadores distinguirem qual a fonte de verdade, quanto mais contribuir.
 
 ![Divergência de IU](/design-systems-for-developers/design-system-inconsistent-buttons.jpg)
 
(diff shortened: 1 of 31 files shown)


content/design-systems-for-developers/react/pt/architecture.md
```diff
@@ -13,7 +13,7 @@ Em grandes empresas, este exercício é feito em conjunção com as equipas de d
...
 
-Nós demos de caras com este problema. Apesar das melhores intenções de uma equipa experiente, os componentes de interface de utilizador eram constantemente reconstruídos, ou colados. Os padrões de IU que deveriam ser os mesmos, começaram a divergir em aparência e funcionalidade. Cada componente tornou-se um floco de neve único, o que fez com que se tornasse impossível para os novos programadores distinguirem qual a fonte de verdade, quanto mais contribuir.
+Nós demos de caras com este problema. Apesar das melhores intenções de uma equipa experiente, os componentes de interface de usuário eram constantemente reconstruídos, ou colados. Os padrões de IU que deveriam ser os mesmos, começaram a divergir em aparência e funcionalidade. Cada componente tornou-se um floco de neve único, o que fez com que se tornasse impossível para os novos programadores distinguirem qual a fonte de verdade, quanto mais contribuir.
```
jonniebigodes: @angelolucas that word choice was applied after quite some research, to avoid "tilting the balance" to one side or another, the word is perfectly valid for both audiences, it might not be commonly used by brazilian readers, but it's valid and both readers can relate to it. The change you're introducing is basically going that way and i personally believe and i'm inclined that  the rest of the team behind Storybook believes the documentation should be generalistic and try to reach as much as a audience as possible. A possible compromise that could applied here, could be change from "utilizador" to "utilizador (ou usuário)".  Sounds good?

jonniebigodes: @angelolucas that word choice was applied after quite some research, to avoid "tilting the balance" to one side or another, the word is perfectly valid for both audiences, it might not be commonly used by brazilian readers, but it's valid and both readers can relate to it. The change you're introducing is basically going that way and i personally believe and i'm inclined that  the rest of the team behind Storybook believes the documentation should be generalistic and try to reach as much as a audience as possible. A possible compromise that could applied here, could be change from "utilizador" to "utilizador (ou usuário)".  Sounds good?

angelolucas: Although correct, we use the term `interface de usuário` here in Brazil. It took me a few seconds to understand the meaning of `interface do utilizador`. I consulted a coworker from Portugal and she confirmed that the most appropriate term is 'usuário'.
 
But okay, you can understand.
Thanks for the review.

jonniebigodes: To add a little more context, when i started the translation effort, i checked with a couple of coworkers of mine in the company i was in, one of them being a brazilian native speaking person and a portuguese teacher friend of mine, this to avoid misleading people and the selection of the wording is correct. With the caveat i alreay mentioned. 

Also i would like to add in the [following](https://dicionario.priberam.org/utilizador) and [also this](https://dicionario.priberam.org/usu%c3%a1rio).  As this change might mislead people, as the usuário has a completely different connotation in portugal, i.e a loan shark. I'm fully aware of the context we're in, but a newcomer that picks up on this might be mislead, hence the proposed compromise

angelolucas: Yes, it really seems to be better to keep the word utilizador.

jonniebigodes:
@domyen and @tmeasday if you both don't mind i'll be closing this pr, based on my comment above. The proposed change would generate some misleadings and some misinformation

domyen:
Close it 👍

Issue #285, opened 2020-02-27
Broken framework links


kjav:
Hi,
I have noticed a few problems with the links at the top of the "get started" pages to versions of the page for other frameworks.

First of all, the React version only links to Vue and Angular, not Svelte or React-native. All of the other pages link to all other pages.

https://learnstorybook.com/intro-to-storybook/react/en/get-started

Secondly, all of the Svelte and React-native links are broken. They link to "/{framework}/{language}/get-started", missing the "/intro-to-storybook" prefix present at the start of the other links. When you manually add this prefix, the links work.

Thanks!

jonniebigodes:
@kjav sorry for any inconvenience that has been caused any inconsistencies in the documentation. But since a couple of releases of Storybook, the documentation has been having some major updates so that not only you, but other readers can get the most accurate information available. Now going by your points:

- Regarding this:

> First of all, the React version only links to Vue and Angular, not Svelte or React-native. All of the other pages link to all other pages.

That is addressed in #283 

Regarding this:

>Secondly, all of the Svelte and React-native links are broken.

The React Native version is also undergoing some changes aswell, as since the time of writing of said documentation and now, some changes happened with both frameworks. A pr addressing the documentation should be up real soon. 

Regarding the links for the Svelte documentation they follow the same structure as the others and should be prepended with the appropriate tutorial prefix, but it looks like it isn't, i've checked and you're right. This will require some further investigation.

kjav:
It's no inconvenience! I just wanted to help & contribute by reporting an issue I found. Thanks for the reply.

jonniebigodes:
@kjav thank you very much for bringing this up. Your contribution is really a valuable one. 

 I've tracked the underlying issue and a pr is up for review.

jonniebigodes:
@kjav this was fixed with #288 if you don't mind i'm going to close the issue

kjav:
Please do, thanks for fixing :)

jonniebigodes:
@kjav no need to thank. Glad that i was able to help you out and i should be the one to thank you for bringing this up.

PR #288, opened 2020-03-02
Feature(site): adds redirects rules for both Svelte and React Native frameworks

jonniebigodes:
This pr introduces 2 new rules to the redirects file in order to fix the navigation issue mentioned in #285. More specifically regarding Svelte and React Native. It prevents 404 when the reader clicks a link that redirects to the tutorial version.

Feel free to provide feedback


tmeasday:
Thanks @jonniebigodes, sorry for the delay

PR #292, opened 2020-03-07
fix: typo on test.md

mrluissan:
"fixed" starts with a non vowel sound, therefore its preceding article should be "a" instead of "an"


tmeasday:
Thanks @mrluissan

PR #294, opened 2020-03-13
Update to correct vue-cli presets

michaelschomburg:
The preset repository `hichroma/vue-preset-learnstorybook` could not be found on github the correct repository for the preset is seen in the commit.

### DIFF
--- a/content/intro-to-storybook/vue/en/get-started.md
+++ b/content/intro-to-storybook/vue/en/get-started.md
@@ -15,7 +15,7 @@ We’ll need to follow a few steps to get the build process set up in your envir
 
 ```bash
 # Create our application, using a preset that contains jest:
-npx -p @vue/cli vue create taskbox --preset hichroma/vue-preset-learnstorybook 
+npx -p @vue/cli vue create taskbox --preset chromaui/vue-preset-learnstorybook 
 
 cd taskbox
 


jonniebigodes:
@michaelschomburg thank you for the contribution. I've just ran a new vue app following the current instructions and it did not throw any errors and the preset was fetched. Or at least the Vue cli did not mention it. The repo you're introducing contains the preset. 

As you can see below:
hichroma-preset-ok

I'm going to defer this to @tmeasday or @domyen for further clarification, just so that we won't be introducing something that could misleads the readers and breaks the initialization process. Are you ok on waiting a bit on this until further clarifications are presented?

michaelschomburg:
@jonniebigodes Yes sure, you are right there though, it does not throw any error. I also only stumbled upon it when I was searching for the preset

jonniebigodes:
@michaelschomburg i'm going to reach out to a core member and see if he can chime in about this and provide more input. Do you mind waiting a bit while this is better looked at?

michaelschomburg:
@jonniebigodes no of-course not, it's not urgent for me at all.

tmeasday:
Yep, we changed the name of the organization a while back and I guess the redirect works, but agree this is better.

PR #299, opened 2020-03-23
WIP adds links to the content and a toc tree

jonniebigodes:
This pr is a WIP to add links to the documentation, more specifically to the headers.

What was done:

- Installed `gatsby-remark-autolink-headers`
- Configured it in `gatsby-config.js`
- Updated the query in `src\components\screens\ChapterScreen\index.js` to use the plugin
- Changed `src\lib\tocEntries` to generate the list of headers based on the data recieved.

What is not yet implemented:
- The urls are yet not fully "sanitized", i've checked and there's some cases with some special characters, more specifically `ç` in the portugues translations and also `ã` in the portuguese and spanish translations. The german as some issues aswell as for the `Ü` case.
- Styling needs to be addressed overall.

I've created a small figma( and pardon me for my lack of skills in it) in [here](https://www.figma.com/file/QVMQZDxH2bJL8yLlh4Ld3G/Concept_for_learnStorybook_link_headers?node-id=0%3A1). 

Even if it doesn't go through like this, i think that at least leaving the plugin `gatsby-remark-autolink-headers` would be a great addition for access to the documentation.

And if a custom svg can be added as the chevron for the links i think it would be a nice touch.

Feel free to provide feedback @domyen @tmeasday @kylesuss

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -26,7 +26,22 @@ module.exports = {
         path: `${__dirname}/content/`,
       },
     },
-    `gatsby-transformer-remark`,
+    {
+      resolve: `gatsby-transformer-remark`,
+      options: {
+        plugins: [
+          {
+            resolve: `gatsby-remark-autolink-headers`,
+            options: {
+              offsetY: 104,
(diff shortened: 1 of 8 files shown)


jonniebigodes:
The build will fail as the node version will need to be changed to >=10 and i'll update the relevant stories shortly

jonniebigodes:
Going to close this out and recreate it with a clean base, more specifically now that #312 was merged. And see if the links can be properly vetted and created. I think it will be doable with `deburr` function that is inside lodash

domyen:
Nice!

**Re: anchor links for header**
They should look something like this. There's a link icon in the design system that outputs an svg, let's reuse it.
image

**Re: Sidebar sub-links**
I like the idea of listing the headings. I don't think we should do it in the sidebar though. There can be a lot of headings per page which will cause the sidebar list to grow beyond what it's designed for. Can we make this a table of contents instead? 

image

PR #303, opened 2020-04-01
stop exporting taskData and actionsData

shabith:
When these 2 item export in storyshots it takes as 2 stories and break the tests.

By the way, what is the real reason to export like this?

### DIFF
--- a/content/intro-to-storybook/react/en/simple-component.md
+++ b/content/intro-to-storybook/react/en/simple-component.md
@@ -59,14 +59,14 @@ export default {
   excludeStories: /.*Data$/,
 };
 
-export const taskData = {
+const taskData = {
   id: '1',
   title: 'Test Task',
   state: 'TASK_INBOX',
   updatedAt: new Date(2018, 0, 1, 9, 0),
 };
 
-export const actionsData = {
+const actionsData = {
   onPinTask: action('onPinTask'),
   onArchiveTask: action('onArchiveTask'),
 };


jonniebigodes:
@shabith thank you for your contribution. Let me go over this and see what might be wrong, but if you have already a repo that i could take a look into it would be great.

To answer this:

> By the way, what is the real reason to export like this?

If you follow along the remainder of the tutorial you'll see that in the next section you'll use data from this section in the next one. More specifically you'll use the `actionsData ` and `taskData`  in there to demonstrate a good practice. 

Feel free to provide feedback

shabith:
Thanks for the quick feedback @jonniebigodes in that case is there anyway we can ignore these exports without building stories?

jonniebigodes:
@shabith that's the reason why the you add the ` excludeStories: /.*Data$/,` as a parameter of the story.  When you issue `yarn storybook` or `npm run storybook`, Storybook  will check the story files you have and as it's starts to process each story it will look up for parameters. If any story has the `excludeStories` element it will ignore any story that matches that particular expression, but use it as a prop to "feed" the story  and component.

If i'm understanding what you want to achieve, you would like for each individual data/action to be a `const` and contained only in each story file so that it would leave each individual story with it's own block of data and actions that relate to that particular story, am i correct?

If that's the case, that's not a bad practice let me tell you and that practice is probably going to find its way into another tutorial that is currently being redone.

But think of it this way, the code as is prevents you from writing the same code over and over again, keeps the stories clean and efficient and indirectly teaches how the exports mechanism works in Javascript as those two snippets will be used across multiple sections and files. I think that's a good thing don't you?

Also personally from a writer's perspective i usually like to keep my snippets as minimal and efficient as possible to keep you (the reader) engaged in the content and avoid you getting sidetracked in one snippet of code trying to understand what it does and with that loose sight of what's being taught.

shabith:
Ah..I'm sorry I missed `excludeStories: /.*Data$/,` part. which create this problem. my bad.

> But think of it this way, the code as is prevents you from writing the same code over and over again, keeps the stories clean and efficient and indirectly teaches how the exports mechanism works in Javascript as those two snippets will be used across multiple sections and files. I think that's a good thing don't you?
Yes you are right. I agree.

Thanks you @jonniebigodes for writing this. It helped me to get better understanding of storybook. 👍

jonniebigodes:
@shabith no need to thank and also no need to appologize aswell. We're all learning here. 

Also, thank @domyen  and @tmeasday as they were the original authors of the tutorial, i just help out fine tuning it as much as i can. 

Also I'm glad that you're getting the grasp of the tutorial and should any issues related to the tutorial feel free to open a issue so that we can help you out as much as we can in providing you with a answer related to the content or improvements to it.

PR #277, opened 2020-02-03
work in progress to address discoverability issue

jonniebigodes:
This pr addresses #270.

This is a work in progress. So far is almost matching the intended changes. There's some styling issues that need to be addressed.

Leaving in the "DO NOT MERGE" label to avoid accidentaly having this merged.

Feel free to provide feedback.

PR #310, opened 2020-04-30
Follow up on #277

jonniebigodes:
This small pr is a follow up on #277. 
Based on the comment [here](https://github.com/chromaui/learnstorybook.com/pull/277#issuecomment-604536270) now the order of the icons/frameworks are corrected. 

As previously this was the resulting in a contradiction with the comment above as you can see below:

storybook_framework_misaligned

On the left storybook running locally and on the right the development preview for ember. And the ember icon is the first one and it shouldn't.

@domyen @kylesuss when you're able, care to take a look? Also let me know if that's the intended order of the frameworks.

Feel free to provide feedback


jonniebigodes:
with the latest commit the `FrameworkContainer` now has a `flex-wap` added to prevent messing up the layout. If that's not the intended behaviour and the frameworks need to stay in one single line i'll update the pr accordingly.

Feel free to provide feedback

domyen:
This is great @jonniebigodes. I was just about to write a comment about flex-wrap based on what's on production :)

jonniebigodes:
@domyen yes, i forgot about that. I'll add it shortly and commit it.

jonniebigodes:
@domyen there you go, just commited it.

domyen:
Is there a way to convert these names into "display" names.

For instance: 
`react-native` 👉 "React Native"
`ember` 👉 "Ember"

jonniebigodes:
we can capitalize the first letter of each framework sure. Let me just see check something and i'll report back.

jonniebigodes:
@domyen just commited what you wanted. I went with a simple approach. Let me know what you think.

domyen:
Awesome work! Merging now

PR #312, opened 2020-05-01
Upgrade Chromatic, Remove segment (add GA, FB pixels separately)

kylesuss:
3 major things going on in here:

1. Remove Segment & individually track certain services (so we don't rely on Segment here)
2. Upgrade to the `chromatic` package
3. Upgrade node versions for our CI process & deployment (Node 8 is outdated, which is what we were on)


gatsby-config.js
```diff
@@ -29,12 +29,26 @@ module.exports = {
...
+    ...(process.env.FACEBOOK_PIXEL_ID && !isDeployPreview
+      ? [
+          {
+            resolve: 'gatsby-plugin-facebook-pixel',
+            options: {
+              pixelId: process.env.FACEBOOK_PIXEL_ID,
+            },
+          },
+        ]
+      : []),
```
kylesuss, line 52: I pulled the env vars into Netlify. No reason to check them in here, I think. I also validated that they work by altering this logic above and removing `!isDeployPreview` to make sure the pixels were there with the right IDs. Now that I know they work, I don't think we should be tracking users on the deploy previews -- only on prod.

.github/workflows/chromatic.yml
```diff
@@ -1,18 +1,17 @@
...
-        token: ${{ secrets.GITHUB_TOKEN }}
-        storybookBuildDir: storybook-static
+      - uses: actions/checkout@v1
+      - run: yarn
+      - run: yarn gatsby telemetry --disable
+      - run: yarn build
+      - run: yarn build-storybook
+      - uses: chromaui/action@v1
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
kylesuss, line 15: This lives in the repo under "Settings" -> "Secrets"

.node-version
```diff
@@ -0,0 +1 @@
+12
```
kylesuss, line 1: Netlify will use this to know which node version to use in the deployment.

.github/workflows/chromatic.yml
domyen: Nice!

PR #314, opened 2020-05-18
Update all code references

tmeasday:
I did a pass over the code repo, updating the react code to:

 - `react-scripts@3.4.1`
 - `react@16.13.1`
 - `@storybook/react@5.3.18`

I also finally updated it to CSF and added commits for the "creating addons" section (based on the text, although it doesn't actually run right now, working on that).

I updated all translations to point to the right place.


jonniebigodes:
@tmeasday regarding the deploy error is due to gatsby graphql layer. The "offending" key is the commit frontmatter key inside the get-started section which is 8741257 and during the build process gatsby detects this as a number and the commit frontmatter key inside for instance the data section is f05981b and it will fetch it as a string. And gatsby does not like it at all. I think that probably it's best for compliance to make all the commits as strings to prevent further issues. 

Also inside `src\components\screens\ChapterScreen\index.js` the commit prop is also being declared as a string so to prevent the issue probably go with stringifying it.

I cloned your fork and did that and the issue went away. Also if my logic is correct the commit frontmatter key is a maintaner one correct? Basically you or me modify it to keep the content updated per the commit revision. Now i know that a reader can directly submit a pr through clicking the link at the bottom of the page in any tutorial. With that probably a good thing is add a note to main readme to inform readers not to change that key as it's for internal consumption. Something like that? 


Feel free to provide feedback

tmeasday:
OK thanks @jonniebigodes - do you want to push your commit to this branch?

jonniebigodes:
@tmeasday there you go. All fixed now. I've just commited the changes and it's working now.

tmeasday:
Appreciate it @jonniebigodes

Issue #275, opened 2020-02-02
Add a banner indicating that a translation isn't updated to latest release


tmeasday:
Based on a discussion in Discord with @stephyswe: 

I think maybe whenever we do a significant release of SB we should add banner to all translations saying "This guide has not been updated yet to SB@X -- if you can help, send a PR here: <>".

Then whenever someone submits a PR to update a translation to that version, that translation can also remove the banner. Probably we'd just have a config field with the version that the translation is known to work with.

WDTY @domyen @jonniebigodes @kylesuss @shilman?

jonniebigodes:
@tmeasday and @stephyswe that is partially addressed, when i did the angular documentation rewording i left in the following in the `get-started` documentation inside a `<div class="aside">`:

> Esta tradução está desatualizada! Ajuda-nos a melhorá-la clicando no link no fim da página. Não somente a equipa agradece, mas toda a comunidade.

Roughly translated to english what is said above is the following: "This translation is not updated! You can help us improve it by clicking the link at the bottom of the page, not only the team will appreciate it but also the entire community"

The same is applied to the spanish version of the said tutorial,

I forgot to create a pull request / create a issue to address this after the changes for the React version went through.

domyen:
Yea, that makes sense to show @tmeasday. We could probably do it as an `.aside` message in Markdown. Good wording @jonniebigodes

kylesuss:
I'm on board w/ the idea. Does seem like it could be hard to remember to add/remove the banner though, especially if it is an all manual process. Anyone have a good idea for that?

jonniebigodes:
@tmeasday @domyen @kylesuss i was thinking about it and the present state of how the documentation is structured. My first thought was to add in a new field by extending the already existing nodes that gatsby will create for the markdown. Something like `lastModified` and possibly use it as a baseline to compare both the original version of the documentation/tutorial with a translation. In paper it sounded good and could probably be done with some effort, but then after some digging it seems that this approach will only work if a contributor or someone willing to make a change to the documentation has the repo cloned locally. Then we could go about it and use a graphql query like the following:
```graphql
markdownRemark(fields: { slug: { eq: $slug } }) {
      
      fields {
        title
      }
      parent {
        ... on File {
          mtime
        }
      }
}
```
[...]

PR #316, opened 2020-05-19
WIP: feature to update the site based on #275

jonniebigodes:
This pr addresses #275. Going to leave in the `DO NOT MERGE` label as this is a work in progress and depending on feedback it will move forward.

What was done:

- Inside `gatsby-config.js` a new element was added to contain the current Storybook version.
- Each tutorials `index.md` frontmatter was updated with a key to contain each individual framework/available language and the version of storybook used in it.
- Created the component to display the information per language if the tutorial is not correct. Probably some available languages, more specifically deutsch/dutch could be better worded.
- Also based on this i kickstarted a i18n approach to the site, for now some parts of the site are done, i'm currently making a inventory of what could be translated in terms of wording component wise. And also if `react-i18next` and other i18n play nice with both Storybook/Gatsby and with that offer a trully translated experience to the reader to his/hers native language, based on what is currently available in terms of content.
- Also updated some of the stories to reflect this change.

Feel free to provide feedback.

### DIFF
--- a/content/design-systems-for-developers/index.md
+++ b/content/design-systems-for-developers/index.md
@@ -64,6 +64,21 @@ contributors:
     },
   ]
 twitterShareText: "I’m learning about building design systems! They're great for scaling frontend code on large teams."
+guideInformation: [
+  {
+    framework: 'react',
+    currentGuideVersion: [
+       {
+         language: 'en',
+         version: 5.3
+       },
+       {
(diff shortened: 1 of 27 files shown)


jonniebigodes:
Going to close this and create a pr later for the issue. I'll need to go back to the drawing board and see how we can better solve it.

PR #326, opened 2020-06-08
Typo fix

razat249:


jonniebigodes:
@razat249 me personally and i believe the rest of the team is pleased with your contribution, but in this case this is not a typo. The wording is correct as when you use the svn command or download the content directly, you'll get a folder with files inside. Those being the fonts and icons required for the styling of the tutorial. The change you're proposing will change the whole sentence, informing the reader that he/she will dowload a folder or files, which is not accurate.

PR #327, opened 2020-06-09
Update get-started.md

deepakkaithwas:
This line in the tutorial was very confusing for me and I was not able to move forward, so I dig into it and came to know the issue.
Instead of making changes in 'App.js' file it should be 'hooks/useCachedResources.js'.

As 'loadResourcesAndDataAsync' function is defined in 'useCachedResources.js' and the purpose of this file is using cached resource.

### DIFF
--- a/content/intro-to-storybook/react-native/en/get-started.md
+++ b/content/intro-to-storybook/react-native/en/get-started.md
@@ -252,10 +252,10 @@ svn export https://github.com/chromaui/learnstorybook-code/branches/master/publi
 svn export https://github.com/google/fonts/trunk/ofl/nunitosans assets/font
 ```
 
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@deepakkaithwas first and foremost, thank you for the submission, when this tutorial was updated, which was early this year if not mistaken, that was the way to go. It seems that between that time and now expo made some changes in their codebase/templates. So with that the code used in this version of the tutorial became a bit outdated. I'm going to check how expo is structuring things now and try it out and if every thing works properly this should be good to go. Do you mind waiting a bit?

Stay safe

jonniebigodes:
@deepakkaithwas sorry for the delay. I'm going to merge this as is. Thank you very much for the time and effort you put into this. 

Stay safe

PR #329, opened 2020-06-12
Import component for typescript

ron4stoppable:
Import location for typescript is different, which is not specified anywhere in document, as mentioned [here](https://github.com/storybookjs/storybook/issues/8817#issue-522091331), I have added a snippet for typescript explicitly.

### DIFF
--- a/content/intro-to-storybook/react/en/using-addons.md
+++ b/content/intro-to-storybook/react/en/using-addons.md
@@ -73,15 +73,10 @@ import { action } from '@storybook/addon-actions';
 import { withKnobs, object } from '@storybook/addon-knobs/react';
 ```
 
-If using typescript
-
-```typescript
-// src/components/Task.stories.tsx
-
-import React from 'react';
-import { action } from '@storybook/addon-actions';
-import { withKnobs, object } from '@storybook/addon-knobs';
-```
+<div class="aside">
+  If you're using TypeScript, you'll need to make a small adjustment to the imports.
+  You'll need to use <code>import { withKnobs, object } from '@storybook/addon-knobs'</code> instead.
+</div>
 
 Next, within the `default` export of `Task.stories.js` file, add `withKnobs` to the `decorators` key:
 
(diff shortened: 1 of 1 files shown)


content/intro-to-storybook/react/en/using-addons.md
```diff
@@ -73,6 +73,16 @@ import { action } from '@storybook/addon-actions';
...
 
+If using typescript
+
+```typescript
+// src/components/Task.stories.tsx
+
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import { withKnobs, object } from '@storybook/addon-knobs';
+```
```
jonniebigodes, line 84: @ron4stoppable instead of adding the snippet we probably could do better and go with the following:

``` html
<div class="aside">If you're using TypeScript, you'll need to make a small adjustment to the imports. You'll need to use <code>import { withKnobs, object } from '@storybook/addon-knobs'</code> instead.</div>
```

Reasoning behind this is to avoid introducing typescript in the middle of the chapter. As the flow and code usage in this version of the tutorial leans towards standard javascript.

jonniebigodes:
@ron4stoppable first and foremost thank you for the contribution. Me personally and i believe the rest of the team is glad that you managed to spot this and made the effort to adjust the documentation. Left a small comment regarding the pr.

Stay safe

content/intro-to-storybook/react/en/using-addons.md
domyen: This is a good compromise.

ron4stoppable: sure thing @jonniebigodes, I was feeling doubtful adding the sample redundancy.

jonniebigodes:
@ron4stoppable sorry for the delay. Once again thank you for the contribution and taking the time to commit the proposed change. I'm going to merge this. 


Stay safe.

ron4stoppable:
Thanks @jonniebigodes for your guidance & merge, happy to  help.

jonniebigodes:
@ron4stoppable no need to thank, glad that i was able to help with this. Also i should be the one to thank you for pointing that small but rather important addition to the documentation.

Stay safe.

PR #331, opened 2020-06-15
Correct grammar in section title and body

chidimo:
Change in heading

continuous deploy --> continuous deployment
Although deploy is the verb form, the adjective "continuous" is the point of emphasis, thus it dictates the structure of the phrase. Thus, the noun, deployment, should follow.

Change in body

continuous deploy --> continuously deploy
In this context, the "deploy" action takes centre stage, thus it dictates that the adverbial form, continuously, precede it.


tmeasday:
Thanks very much

PR #336, opened 2020-06-19
Fix letter case and wrong file path on tutorial

jparksecurity:
Because of letter case, style doesn't properly work, so I fix it.

And I try to be constant with file path


jonniebigodes:
@jparksecurity going to merge this. Thank you for spotting this. 

Stay safe

PR #246, opened 2020-01-07
Chore: Bump packages versions.

jonniebigodes:
With this pr the packages versions with the exception of formik were updated.

Moving `formik` to the latest version, introduces a bug with not only the stories but also with the site. I haven't pinpointed it yet, so i thought it would be best to leave it in the current version to avoid breaking functionality.

Also contributing to the effort of changing the stories format to csf. As the documentation is being moved, also a good thing would be that learnstorybook's site would benefit from this aswell.

I'm adding the label `DO NOT MERGE` temporarily as this is reviewed in more detail

Feel free to provide feedback

Issue #225, opened 2019-12-04
Update react code + commits


tmeasday:
With some recent changes:

## Intro to SB
 - [ ] Latest SB version
 - [ ] CSF
 - [ ] ?

## DS for devs
 - [ ] Latest SB version
 - [ ] CSF
 - [ ] Update preset: https://github.com/chromaui/learnstorybook.com/pull/187/files

PR #339, opened 2020-06-24
WIP: Move LearnStorybook to Storybook version 6.0

jonniebigodes:
This pr addresses #225, the Storybook version and stories inside were updated to version 6.0.

What was done:

- Updated the Storybook packages to the latest 6.0 version, which is `6.0.0-beta.34`.
- Updated the Storybook design system package to the latest to prevent the issue that presented itself in [here](https://github.com/chromaui/learnstorybook.com/pull/246#issuecomment-575696088).
- Updated the stories to reflect these changes. While preserving the original component story name.

One thing that is happening to me and i would like for feedback on this. Is when i start storybook locally, the following is happening:

lsb_sb_60_1

As you can see the path is defaulting to null and no story is being loaded. Am i missing something here?

@tmeasday and @domyen and @kylesuss feel free to provide feedback.


jonniebigodes:
Hum...it seems that the issue that i've mentioned in the description is presenting itself once again. Going to check if it could be related to the `styled-components` package being used in the design system and the version being used in the site,

domyen:
Hey, re:undefined path. I encountered that yesterday as well. I think
@tmeasday and @ndelangen have a fix in mind. Not sure it’s been released
yet though.

jonniebigodes:
@domyen i'll be on the lookout and see check the releases. A new version was released, but according with the it, there's no mention to this. So probably it hasn't made it's way into the package.

src/components/basics/GatsbyLink.stories.js
```diff
@@ -1,7 +1,10 @@
...
 
-storiesOf('Basics|GatsbyLink', module)
-  .addParameters({ component: GatsbyLink })
-  .add('default', () => <GatsbyLink to="/">I am a GatsbyLink</GatsbyLink>);
+export default {
+  component: GatsbyLink,
+  title: 'Basics/GatsbyLink',
+};
+
+export const Basic = () => <GatsbyLink to="/">I am a GatsbyLink</GatsbyLink>;
```
shilman, line 9:
```suggestion
export const Default = () => <GatsbyLink to="/">I am a GatsbyLink</GatsbyLink>;
```

Suggest don't use `storyName` when not needed for cleanliness

src/components/basics/GatsbyLink.stories.js
```diff
@@ -1,7 +1,10 @@
...
+Basic.storyName = 'Default';
```
shilman, line 10:
```suggestion
```

src/components/composite/Header.stories.js
```diff
@@ -29,11 +28,35 @@ const props = {
...
-    </BlueBackground>
-  ));
+export default {
+  component: Header,
+  title: 'Composite/Header',
+};
+
+export const Basic = args => <Header {...args} />;
+Basic.storyName = 'default';
+Basic.args = {
```
shilman, line 38: Redundant with `props`?

src/components/composite/Header.stories.js
```diff
@@ -29,11 +28,35 @@ const props = {
...
+          },
+          fields: {
+            slug: 'guide-slug',
+          },
+        },
+      },
+    ],
+  },
+};
+export const Inverted = () => (
```
shilman, line 57: convert to Args format and spread in `Basic.args`?

src/components/screens/ChapterScreen/Pagination.stories.js
```diff
@@ -1,14 +1,20 @@
...
 };
 
-storiesOf('Screens|ChapterScreen/Pagination', module)
-  .addParameters({ component: Pagination })
-  .add('without nextEntry', () => <Pagination />)
-  .add('with nextEntry', () => <Pagination nextEntry={nextEntry} />);
+export const Basic = args => <Pagination {...args} />;
+Basic.storyName = 'without nextEntry';
+
+export const WithNextEntry = Basic.bind();
```
shilman, line 12: We're recommending the following syntax:

```js
const PaginationStory = (args) => <Pagination {...args} />;

export const Basic = PaginationStory.bind({});
Basic.args = { ... };

export const WithNextEntry = PaginationStory.bind({});
WithNextEntry.args = { ... };
```

tmeasday:
@jonniebigodes @shilman just releases `beta.36` which fixes the above issue.

src/components/screens/GuideScreen/Hero.stories.js
```diff
@@ -89,53 +94,95 @@ const storyLanguages = [
...
+  title: 'Visual Testing Handbook',
+  contributorCount: '34+',
+  languages: defaultLanguagesData,
+};
+export const WithChapterCount = Basic.bind();
+WithChapterCount.storyName = 'with chapter count';
+WithChapterCount.args = {
+  ctaHref: '/get-started',
+  description:
+    "Visual testing is a pragmatic yet precise way to verify the look of UI components. It's practiced by companies like Slack, Lonely Planet and Walmart. This 31-page handbook gives you an overview of visual testing in React.",
```
tmeasday, line 129: If you want to reuse an arg across stories the easiest thing to do it add it to the component's args:

```js
export default {
 ...
 args: {
   description: "...",
  }
}
```

It will apply to every story.

If you want to only use the same arg for a subset of stories, you can override it for the stories where you don't, OR you can simply re-use the JS object:

```js
Basic.args = {
  // ...
 description: "...",
};

// Just specify the new stuff. 
WithContributorCount.args = {
   ...Basic.args,
   contributorCount: '34+',
}
```

jonniebigodes:
with the latest commit the stories were updated per feedback and aswell the packages version. @shilman and @tmeasday if you're able to take another look at this and see if that's ok, then this should be finally be good to go

domyen:
lgtm tests approved

tmeasday:
Thanks @jonniebigodes!

Have been talking through some of these stories with @shilman and we think it's important to get the patterns right so we set an example to ourselves and others going forward. Sorry if it seems we are being inconsistent, we want to try and nail this down!

Also really interested in what you thing about the developer ergonomics of all this having "written" a bunch of stories in the args format now.

Some notes:

### Use of component-level args.

It feels like in some cases you've put args on the component (especially when there is only one story for the component) where they feel quite specific to the test case of that story. For instance:

https://github.com/chromaui/learnstorybook.com/blob/47e6a12ccb37bb7a429601a012032a9d47c27658/src/components/composite/ShadowBoxCTA.stories.js#L16-L27

Here it feels like the heading and message at least feel pretty specific.

In other cases the component-level args feel appropriate as the args aren't that interesting and are just there to get the component to render, e.g.

https://github.com/chromaui/learnstorybook.com/blob/47e6a12ccb37bb7a429601a012032a9d47c27658/src/components/screens/GuideScreen/Hero.stories.js#L8-L15

I'm not sure there's a hard-and-fast rule here, what do you think?

We were sort of thinking it might go something (if you were writing stories):

1. Write your first story, include all args on the story
2. Write your second story, if you need to reuse the args, use a spread, add some changes.
3. Write your third story, if you feel like you are spreading the same things again, consider using component-level args.

WDYT?

### Use of `Story` component

I noticed that in the story files with only one story you've not created a `Story` component and cloned (bind-ed) it.

I wonder if we should encourage always creating a component and binding, just for consistency. WDYT?

PS. @shilman suggested he always calls such components `Story` (so it's `Story.bind({})`) rather than e.g. `ButtonStory()`. This may help with tooling and in a way reads better.. `const LoggedIn = Story.bind({})`, sort of reads "LoggedIn is a story".

### Reuse of args

I think there are a few opportunties for reuse both within and across story files now that we are using args (yay!). For instance these args:

https://github.com/chromaui/learnstorybook.com/blob/47e6a12ccb37bb7a429601a012032a9d47c27658/src/components/screens/ChapterScreen/LanguageMenu/index.stories.js#L27-L37

Could be reused (spread into) the subsequent stories.

Or these args:

https://github.com/chromaui/learnstorybook.com/blob/47e6a12ccb37bb7a429601a012032a9d47c27658/src/components/screens/GuideScreen/Contributors.stories.js#L7-L22

Could be reused in this story:

https://github.com/chromaui/learnstorybook.com/blob/47e6a12ccb37bb7a429601a012032a9d47c27658/src/components/screens/GuideScreen/index.stories.js#L12-L25

jonniebigodes:
@tmeasday i'm going through your feedback and it makes sense to me. The main reason i moved the args to be global rather than storywise was to allow it to be expanded, should the stories for that case expand, at least we would have a baseline data available across all of the stories. But i like you pointed it out. Going with :

> We were sort of thinking it might go something (if you were writing stories):

>1- Write your first story, include all args on the story
>2- Write your second story, if you need to reuse the args, use a spread, add some changes.
> 3- Write your third story, if you feel like you are spreading the same things again, consider using component-level args.

I think it's a way better approach. 

The `Hero.stories.js` applies that principle nicely, we have more than 3 stories. Common data that is relevant to all stories, so it's pushed to the let's call it "component args" (pardon the bad analogy), also we're reusing args that are in one story but are needed in the another story alongside with the args that are relevant to the actual story.

Regarding the `ShadowBoxCTA` to be perfectly honest, i was testing something, i was checking if the args would allow markup inside a arg and if it would behave. I was going through the RFC and couldn't find a definite answer to that question,or i've missed it, so and based on my testing it worked and left it as it could be a good way to demonstrate that it can be used.

Also other cases like that one, where there's only one story i have no issues with going with the `.bind({])` approach. For instance `BoxLink.stories.js` could be turned into:

```js
const Story = args => <BoxLink {...args}>BoxLink content</BoxLink>;
export const Default = Story.bind({});
Default.args = {
  to: '/',
};
```

I think that applying this principle to the other stories aswell enforces consistency and a good way to setup the stories in this case and for further use cases. 

Regarding this:

>PS. @shilman suggested he always calls such components Story (so it's Story.bind({})) rather than e.g. ButtonStory(). This may help with tooling and in a way reads better.. const LoggedIn = Story.bind({}), sort of reads "LoggedIn is a story".

I took a note regarding this. i was actually planning on going with that, calling it `Story` and afterwards binding it, but i wasn't sure if it would work. But it seems that it will work. So i'll make the change and give it a go.

Regarding this:
> Have been talking through some of these stories with @shilman and we think it's important to get the patterns right so we set an example to ourselves and others going forward. Sorry if it seems we are being inconsistent, we want to try and nail this down!

My thoughts exactly and seriously no need to appologize whatsoever.

jonniebigodes:
@tmeasday and @shilman i've just made some changes to the stories files, per feedback. I would like for you to both to take a look at the `src\components\composite\Header.stories.js` file, how it's currently implemented and how it's implemented in this pr, basically how to address the decorators on a story. Also i believe there's one or more cases that have the same approach. 

Regarding `learnstorybook.com/src/components/screens/ChapterScreen/LanguageMenu/index.stories.js` i'm going to take a look at it and see how the data can be extracted and shared across some other stories that use it shortly and then commit the changes and we'll circle back to that particular case. In the meantime if you can take another look and see if you're ok with it, i believe we're in good shape and almost have a good baseline to be used here, but also a good baseline for other developers aswell.

Feel free to provide feedback

src/components/composite/Header.stories.js
```diff
@@ -29,11 +35,9 @@ const props = {
...
-    <BlueBackground>
-      <Header isInverted {...props} />
-    </BlueBackground>
-  ));
+export const Inverted = Story.bind({});
+Inverted.args = {
+  ...Default.args,
+  isInverted: true,
+};
+Inverted.decorators = [story => <BlueBackground>{story()}</BlueBackground>];
```
tmeasday, line 43: Yeah I am a big fan of this 👍 (the decorator).

jonniebigodes: You and me. This way of organizing the stories and bringing in items like decorators closer to the story feels more organic to me.

src/components/composite/Header.stories.js
jonniebigodes: This was fixed in a prior commit. Thanks for the time reviewing it.

src/components/composite/Header.stories.js
jonniebigodes: This was fixed in a prior commit. Thanks for the time reviewing it.

src/components/screens/ChapterScreen/Pagination.stories.js
jonniebigodes: This was fixed in a prior commit. Thanks for the time reviewing it.

src/components/basics/GatsbyLink.stories.js
jonniebigodes: This was fixed in a prior commit. Thanks for the time reviewing it.

domyen:
Can you fix this story for [IndexPage](https://www.chromatic.com/snapshot?appId=5cf841a3f3e3d200208ffc74&id=5ef9f807e844ba0022693744)? It seems to be a regression.

Let's merge it after the fix 👍

PR #340, opened 2020-06-24
adding curl and tar as option to add assets

aaroncalderon:
I stumbled upon the issue that I did not had `svn` installed, and I prefer a command rather than manually downloading and then moving the folders.

So, I researched a way to use `curl` instead.

I hope you find this pull request helpful and that you merge it.

### DIFF
--- a/content/intro-to-storybook/react/en/get-started.md
+++ b/content/intro-to-storybook/react/en/get-started.md
@@ -73,4 +73,14 @@ svn export https://github.com/chromaui/learnstorybook-code/branches/master/publi
 svn export https://github.com/chromaui/learnstorybook-code/branches/master/public/font public/font
 ```
 
+<div class="aside">
+  <p>Additionally, if you prefer <code>curl</code> and <code>tar</code> to download a folder of files from GitHub instead use the below command.</p></div>
+  
+```bash
+curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master --output c.gz 
+tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/font 
+tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/icon 
+rm c.gz
+```
+
 After adding styling and assets, the app will render a bit strangely. That’s OK. We aren’t working on the app right now. We’re starting off with building our first component!
--- a/content/intro-to-storybook/react/en/get-started.md
+++ b/content/intro-to-storybook/react/en/get-started.md
@@ -66,21 +66,14 @@ If you want to modify the styling, the source LESS files are provided in the Git
 To match the intended design, you'll need to download both the font and icon directories and place its contents inside your `public` folder.
 
 <div class="aside">
-<p>We’ve used <code>svn</code> (Subversion) to easily download a folder of files from GitHub. If you don’t have subversion installed or want to just do it manually, you can grab the folders directly <a href="https://github.com/chromaui/learnstorybook-code/tree/master/public">here</a>.</p></div>
-
-```bash
-svn export https://github.com/chromaui/learnstorybook-code/branches/master/public/icon public/icon
-svn export https://github.com/chromaui/learnstorybook-code/branches/master/public/font public/font
-```
-
-<div class="aside">
-  <p>Additionally, if you prefer <code>curl</code> and <code>tar</code> to download a folder of files from GitHub instead use the below command.</p></div>
+<p>We’ve used curl and tar to easily get the most recent assets from GitHub. If you don’t have curl installed or want to just do it manually, you can grab the folders directly <a href="https://github.com/chromaui/learnstorybook-code/tree/master/public">here</a>.</p>
+<p>Depending on your OS, you probably need to update some of the commands to allow the assets to be copied without any issues.</p>
+</div>
   
 ```bash
-curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master --output c.gz 
-tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/font 
-tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/icon 
-rm c.gz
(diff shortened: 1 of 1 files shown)


domyen:
Thanks @aaroncalderon. Appreciate the PR! 🙇 

I think every machine should be able to do this without installing anything extra.

@jonniebigodes @tmeasday should we make `curl` the default way to get these folders?

jonniebigodes:
@domyen if not mistaken, curl is added in Windows 10, upwards version 1803 or later. And also if not mistaken, if the reader has installed git for windows it will add curl aswell. So in that department we're good, i have no issue. My only issue is where to fetch it and extract it. I have to test this out first.

tmeasday:
Yeah if curl works it is probably a lot less confusing that `svn`. Can we do it all in one line? `curl ... | tar -zxf -` or something?

domyen:
Let's switch the commands and remove svn altogether. Would also prefer if we can do it in one line. Is that possible @aaroncalderon?

arunoda:
@tmeasday It seems like GitHub already supports git's sparse-checkout.
May be consider using it.

Ref: https://twitter.com/mnmtanish/status/1276174713572610048

content/intro-to-storybook/react/en/get-started.md
```diff
@@ -73,4 +73,14 @@ svn export https://github.com/chromaui/learnstorybook-code/branches/master/publi
...
 
+<div class="aside">
+  <p>Additionally, if you prefer <code>curl</code> and <code>tar</code> to download a folder of files from GitHub instead use the below command.</p></div>
+  
+```bash
+curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master --output c.gz 
+tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/font 
+tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/icon 
+rm c.gz
+```
```
jonniebigodes, line 84: @aaroncalderon @domyen  and @tmeasday. I was checking and it looks like we can do something like:

```bash
curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master --output assets.gz

tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/font | tar -C public -xvzf c.gz --strip=2 learnstorybook-code-master/public/icon

rm assets.gz
```

Text wise we could change it to:

```html
<div class="aside">
<p>We’ve used curl and tar to easily get the most recent assets from GitHub. If you don’t have curl installed or want to just do it manually, you can grab the folders directly here (<- the here link stays the same).</p>
<p>Depending on your OS, you probably need to update some of the commands to allow the assets to be copied without any issues.</p>
</div>
```


Probably something like this? 

And we could drop the svn alltogether.

Feel free to provide feedback

tmeasday: I think this oneliner works: 

```
curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master  | tar -zxvC public --strip=2 learnstorybook-code-master/public/{font,icon}
```

WDYT @jonniebigodes?

tmeasday:
👋 @arunoda 

> @tmeasday It seems like GitHub already supports git's sparse-checkout.
> May be consider using it.

Looks like you need git 2.25 to use that and at least on my machine I have 2.13. Ideally we wouldn't ask people to install anything.

tmeasday:
@jonniebigodes IIRC there are lots of these `svn` commands through the design systems chapters. If we can figure out a good pattern here we should change all of them too.

content/intro-to-storybook/react/en/get-started.md
jonniebigodes: @tmeasday let me check it out and get back to you.

jonniebigodes:
@tmeasday doing a quick check in the "Design Systems for Developers" guide the "Architecture" chapter is the only one that uses svn. I'm going to check more thoroughly, but it seems that it's the only place.

content/intro-to-storybook/react/en/get-started.md
jonniebigodes: @tmeasday did a run with it and what you posted works fine on a WSL, meaning probably all the NIX based should be good to go. So i think that with the adjustment to this command we can streamline and remove svn in here. Also if the reader is running Windows we've covered that mentioning that he needs to adjust the command.

tmeasday: Ok, sounds good to me @jonniebigodes

jonniebigodes: @aaroncalderon when you're able, can you address the change and commit it, so that we can this merged?

aaroncalderon: got it...

Working on it.

aaroncalderon: > curl https://codeload.github.com/chromaui/learnstorybook-code/tar.gz/master  | tar -zxvC public --strip=2 learnstorybook-code-master/public/{font,icon}

@tmeasday I had issues with piping the output of `curl` to `tar`. That is why I split the command so I would get a file, then i would use the file to extract the desired files and finally remove the temporary file.

I would like to keep the command in separate lines. Although that one liner looks neat.

aaroncalderon:
> Let's switch the commands and remove svn altogether. Would also prefer if we can do it in one line. Is that possible @aaroncalderon?

@domyen I tried to make it a one liner, however, I had issues with piping the output of `curl` to `tar`. Something about `tar` not waiting for the stream to finish, and then `curl` throwing an error.

So, I had to use a temporary file and separate line commands. That is why I made it multiple lines.

jonniebigodes:
@aaroncalderon with #370 we went with a bit of different approach and used [degit](https://github.com/Rich-Harris/degit) as it allows a simpler and easier way to get folders from GitHub. 

Nonetheless i would like to thank you for the time and effort you put into the pull request and kickstarting this update. We're truly grateful for it.

Let us know if there are any issues and we can improve the documentation even more.

Stay safe

PR #355, opened 2020-08-15
Fix git commands

gregrickaby:
When I copied & pasted these two commands (on my MacBook Pro) I received the following errors:

`git add src\components\Task.js`

```bash
fatal: pathspec 'srccomponentsTask.js' did not match any files
```

`git commit -m “change task background to red“`

```bash
error: pathspec 'task' did not match any file(s) known to git
error: pathspec 'background' did not match any file(s) known to git
error: pathspec 'to' did not match any file(s) known to git
error: pathspec 'red”' did not match any file(s) known to git
```

### DIFF
--- a/content/intro-to-storybook/react/en/test.md
+++ b/content/intro-to-storybook/react/en/test.md
@@ -64,13 +64,13 @@ This yields a new background color for the item.
 Add the file:
 
 ```bash
-git add src\components\Task.js
+git add src/components/Task.js
 ```
 
 Commit it:
 
 ```bash
-git commit -m “change task background to red”
+git commit -m "change task background to red"
 ```
 
 And push the changes to the remote repo:
--- a/content/intro-to-storybook/react/en/test.md
+++ b/content/intro-to-storybook/react/en/test.md
@@ -64,7 +64,7 @@ This yields a new background color for the item.
 Add the file:
 
 ```bash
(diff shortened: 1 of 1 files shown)


domyen:
Yikes, thanks for catching this. I must've missed it! 

Would you mind double-checking if this is the case in the other translations as well? That way we can fix it for every reader. 🙇‍♂️

gregrickaby:
@domyen 🤔 a find/replace doesn't show any other results.

Thumbing through each translated `test.md`, there inconsistencies with the git commands throughout. 

The `git add` command for example:

- `/react/de/test.md` uses `$ git add -A`
- `/react/es/test.md` uses `$ git add .`

The inconsistencies occur with the `git commit` command as well, and most translations just use: `$ git commit -m "taskbox UI"`

Would you like me to update the English translation to match the other translations?

jonniebigodes:
@gregrickaby i was the one that updated the documentation. And when was modifying it i working with a windows box. If you're ok and to avoid confusion, let's go with the `git add .` instead of what's currently documented. 

Let me know when you make the change so that i can merge.

Stay safe

gregrickaby:
@jonniebigodes All set!

jonniebigodes:
@gregrickaby thank you for the speedy change. Going to merge it

PR #368, opened 2020-08-27
Tweak subtitle for brevity

domyen:
Minor tweak


jonniebigodes (approved):
I like it. Reads way better. Going to merge it now

PR #369, opened 2020-08-27
Removes reference to SVN because it is no longer 1999

obeid:
Because this shit fucked with me real hard, and on the first page of your otherwise very nice book.
Like why? 

https://www.youtube.com/watch?v=shbxUe-S2V8

### DIFF
--- a/content/intro-to-storybook/react/en/get-started.md
+++ b/content/intro-to-storybook/react/en/get-started.md
@@ -65,12 +65,9 @@ If you want to modify the styling, the source LESS files are provided in the Git
 
 To match the intended design, you'll need to download both the font and icon directories and place its contents inside your `public` folder.
 
-<div class="aside">
-<p>We’ve used <code>svn</code> (Subversion) to easily download a folder of files from GitHub. If you don’t have subversion installed or want to just do it manually, you can grab the folders directly <a href="https://github.com/chromaui/learnstorybook-code/tree/master/public">here</a>.</p></div>
-
 ```bash
-svn export https://github.com/chromaui/learnstorybook-code/branches/master/public/icon public/icon
-svn export https://github.com/chromaui/learnstorybook-code/branches/master/public/font public/font
+npx degit https://github.com/chromaui/learnstorybook-code/branches/master/public/icon public/icon
+npx degit https://github.com/chromaui/learnstorybook-code/branches/master/public/font public/font
 ```
 
 After adding styling and assets, the app will render a bit strangely. That’s OK. We aren’t working on the app right now. We’re starting off with building our first component!


obeid:
1999 was a Prince reference, but then I realized Subversion was actually made in 1999. 👍

domyen:
Wow, I did not know about `degit`. We have an [open PR](https://github.com/chromaui/learnstorybook.com/pull/340) to swap out svn with curl. This seems even easier. What do you think @jonniebigodes?

jonniebigodes:
@obeid thank you for the time and effort you put into this. I'm going to take a look at it and see if works with all the frameworks being used with the tutorial. Should that be the case we could use the `degit` package instead and we adjust the wording (the aside that was removed) so that the reader can get a sense of what he/she is doing. 

Also, on a maintainer's perspective if you don't mind please update the pull request description and title, also avoid using profanities. It doesn't sit well with us. I personally liked the Prince reference (brings me back to my late teens :P), but other people and maintainers might not see it that way.

Sounds good?

jonniebigodes:
@obeid thank you for the time and effort you put into this. I'm going to take a look at it and see if works with all the frameworks being used with the tutorial. Should that be the case we could use the `degit` package instead and we adjust the wording (the aside that was removed) so that the reader can get a sense of what he/she is doing. 

Also, on a maintainer's perspective if you don't mind please update the pull request description and title, also avoid using profanities. It doesn't sit well with us. I personally liked the Prince reference (brings me back to my late teens :P), but other people and maintainers might not see it that way.

Sounds good?

obeid:
@jonniebigodes To _fuck with_ something is not the same as to _fuck_ said thing. 

The aside seemed to exist mainly to justify the use of Subversion, the user already got a sense of what they are doing from the previous sentence. 

I have turned off allowing edits by maintainers. Take my free labor as it is, or leave it.

obeid:
FWIW `curl` doesn't come pre-installed on Ubuntu, arguably the most popular Linux distribution out there.

jonniebigodes:
@obeid regarding your last comment. Hence why it hasn't been merged. I had on my todos to come up with a better way.

On the topic at hand.

What you're introducing won't work as is. I just tested it on a fresh project.

If you issue this:

```bash
npx degit https://github.com/chromaui/learnstorybook-code/branches/master/public/icon public/icon
```
It clones only the folder as you can see below, this is the result of your proposed change.

degit-finder

Vscode opened will reflect the behavior and show me the following

degit-vscode


Checking the documentation for [degit](https://github.com/Rich-Harris/degit#specify-a-subdirectory) what needs to be used is:

```bash
npx degit chromaui/learnstorybook-code/public/font public/font 
```

This will yeld the following result:

degit-working-vscode
finder-degit-working

If you're willing to make the changes and once i vet if it works for the other frameworks this will be merged.

Regarding your other comment. I stand for what i've said, i understand what you're meaning, but others don't hence why i mentioned it.

Stay safe

jonniebigodes:
@obeid Going to close this in favor of #370. The documentation is now updated globally to use degit. 

Nonetheless thank you for the time and effort you put into this pull request and offer a better way to pull in the files/folders and improve the documentation.

We really appreciate it.

Stay safe.

PR #372, opened 2020-08-28
Changed capitalization when using Default property

MatsDahlin:
The jest test was broken because of the lower case 'd' when accessing the Default property of the imported TaskStories.


jonniebigodes:
@MatsDahlin absolutely correct. Sorry about that. Copied over the wrong snippet when i was updating the docs. Going to merge it.

Stay safe

PR #376, opened 2020-09-01
Update Svelte simple-component.md

jdhines:
The double brackets in the class caused me to miscode it the first time such that the pinned state wasn't reflected.


jonniebigodes:
@jdhines let me take a look, but i believe the current documentation is correct. If you don't use a string template the css class per state won't be applied. With what you're proposing it might not be triggered.

Do you mind waiting a bit?

Stay safe.

jonniebigodes:
I stand corrected on my earlier comment. Tested it on a fresh svelte project with Storybook 6 and the proposed change works like a charm in either development or production. 

Thank you for taking the time to make this pull request and optimizing the documentation. Also sorry for not addressing this sooner, but i had to take care of some other items that needed my attention.

Going to merge it @jdhines 


Stay safe.

PR #381, opened 2020-09-07
change yarn start to yarn web

dannyhw:
Running yarn start will only run the bundler and not the app.

I've added a comment here to let people know that the stories won't actually load at this point since this is causing some confusion. It would also be good to update the screenshots here.

In order for the stories to be loaded on the serverUI like they are in the screenshot the storybookUI needs to be loaded. This happens later when the linksScreen gets updated.

### DIFF
--- a/content/intro-to-storybook/react-native/en/get-started.md
+++ b/content/intro-to-storybook/react-native/en/get-started.md
@@ -89,7 +89,7 @@ yarn test
 yarn storybook
 
 # Run the frontend app proper on port 19002:
-yarn start
+yarn web
 ```
 
 ![3 modalities](/intro-to-storybook/app-three-modalities-rn.png)
--- a/content/intro-to-storybook/react-native/en/get-started.md
+++ b/content/intro-to-storybook/react-native/en/get-started.md
@@ -94,6 +94,8 @@ yarn web
 
 ![3 modalities](/intro-to-storybook/app-three-modalities-rn.png)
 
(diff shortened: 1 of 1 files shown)


content/intro-to-storybook/react-native/en/get-started.md
```diff
@@ -94,6 +94,8 @@ yarn web
 
 ![3 modalities](/intro-to-storybook/app-three-modalities-rn.png)
 
+Your storybook server won't be able to load the stories list just yet, but we'll fix this in the next steps.
+
```
jonniebigodes, line 98: @dannyhw we could probably improve this by going with:

"Checking our Storybook at this point, you might see that there's no stories displayed. That's ok, we'll take care of it shortly, for now, let's continue working on getting our application properly setup."

Something in like this

Let me know if you're ok with it

jonniebigodes (changes requested):
@dannyhw I've left a comment regarding the wording, we probably can go with something like this.

content/intro-to-storybook/react-native/en/get-started.md
dannyhw: That seems good to me, I'll update it to use your wording 👍

dannyhw: its updated

jonniebigodes:
@dannyhw thank you for the time and effort you put into this and help fine tune the tutorial. I'm going to merge it.

Stay safe

PR #383, opened 2020-09-16
Suggest creating new branch command in test.md

juunone:
[github.blog/2019-08-16](https://github.blog/2019-08-16-highlights-from-git-2-23/). 
Git 2.23 brings a new pair of commands `git switch` and `git restore`
how about change new branch command with `git switch`?

Feel free to provide feedback
Thank you.

### DIFF
--- a/content/intro-to-storybook/react/en/test.md
+++ b/content/intro-to-storybook/react/en/test.md
@@ -39,7 +39,7 @@ Let's see how it works by tweaking the background of the `Task` component.
 Start by creating a new branch for this change:
 
 ```bash
-git checkout -b change-task-background
+git switch -c change-task-background
 ```
 
 Change `Task` to the following:


jonniebigodes:
@juunone let me take a closer look at this and get back to you. This could probably be added to the rest of the documentation.

domyen:
Thanks @juunone This seems like a pretty new git feature afaict. How can we tell if peoples machines are updated to take advantage of this (esp windows)?

juunone:
@jonniebigodes Thanks for feedback.
rest of the documentation change is very helpful!

juunone:
@domyen This feature is support git version 2.23.0 or higher
window and mac both of them check with `git --version` command.

I hope my answer was helpful.
thank you.

content/intro-to-storybook/react/en/test.md
```diff
@@ -39,7 +39,7 @@ Let's see how it works by tweaking the background of the `Task` component.
 Start by creating a new branch for this change:
 
 ```bash
-git checkout -b change-task-background
+git switch -c change-task-background
 ```
```
jonniebigodes, line 43: @juunone we probably can reach a bit of a compromise here. Instead of shifting this to what you're proposing, we probably add the switch command as a aside. Turning this into:

```md
Start by creating a new branch for this change:


git checkout -b change-task-background

<div class="aside">
If you're using a newer version of git, you can replace this command with <code>git switch -c change-task-background</code>
</div>
````

The reason behind this, is that we cannot know for sure when the reader has installed git or if it's keeping it updated and if we make this as the instruction to be used, people might run into issues.

To be honest with you my windows rig was running a outdated version for quite some time and during that time i contributed to this and other repos and didn't even bother to check for updates (i'm aware that it's a bad practice), it was up until recently that i've updated to a newer version.

Let me know if you're ok with it and we can include this into the other frameworks covered by the tutorial.

jonniebigodes (changes requested):
@juunone Thank you for the time and effort you put into the pull request and help us improve it. I left a comment and would like for you to see if you're ok with it and we'll move from that.

Stay safe

domyen:
Thanks for testing it @jonniebigodes. It doesn't seem like this is going to work 100% of the time. I don't think we can recommend `switch` because it wont work for readers on older git versions and generate more issues for maintainers (us) in the future. The compromise ends up making the guide longer.

Closing this issue for now.

content/intro-to-storybook/react/en/test.md
juunone: @jonniebigodes Thanks for detail answer!
you right, maybe not all of people check for git version.
it wont work for readers on older git versions so..
your turning guide answer is satisfies for everyone.

thank you for your time and hope to it incldue other tutorials.
thank you

PR #384, opened 2020-09-19
missing closing tag

cope:
adding missing </script> closing tag


cope:
not really needed after second thought...

PR #388, opened 2020-09-25
Update using-addons.md

icruzr93:
Minimal copy fix to change "a addon" to "an addon"


domyen (approved):
thanks for the grammar fix! 🙏

PR #391, opened 2020-10-01
Update simple-component.md

fehrm:
For Angular, addon-notes isn't included in the default configuration, so you get an error from NPM about a missing package when restarting Storybook after the change.

Simple fix, remove addon-notes from the addons array.

### DIFF
--- a/content/intro-to-storybook/angular/en/simple-component.md
+++ b/content/intro-to-storybook/angular/en/simple-component.md
@@ -147,7 +147,7 @@ We also need to make one small change to the Storybook configuration so it notic
 // .storybook/main.js
 module.exports = {
   stories: ['../src/app/components/**/*.stories.ts'],
-  addons: ['@storybook/addon-actions', '@storybook/addon-links', '@storybook/addon-notes'],
+  addons: ['@storybook/addon-actions', '@storybook/addon-links'],
 };
 ```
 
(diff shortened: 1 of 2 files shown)


jonniebigodes:
@fehrm thank you for the time and effort you put into this pull request. As you may or may not know we released Storybook version 6.0 recently and with that the contents of this version of the tutorial is now a bit outdated. If you're willing we would love if you could help us with an update to the tutorial. 

But for now, if you could remove the reference from the portuguese translation as well we would appreciate it and i'll gladly merge this. 

Let me know once you've updated it.

Stay safe

fehrm:
Hi @jonniebigodes 
Thanks for spotting the mistake! 

I did had to do some small changes regarding the package versions etc to get everything working. I can try to create a new PR during the weekend with my changes if that would help you out?

jonniebigodes:
@fehrm don't worry about it. No problem whatsoever, we more than welcome the help you're willing to provide. If you're ok and create a repo for a reproduction like setup in [here](https://github.com/jonniebigodes/learnstorybook-v6-vue) and go about it like it's in there it would be fantastic. You would break up the code into stages and we could go over it as we're progressing and in cause we hit some "road bumps" (pardon the pun) we could better triage them.

Let me know when you got a chance to setup the repo.

On a side note, we're thrilled that you decided to learn about Storybook and are willing to help us with the tutorial

PR #397, opened 2020-10-11
Update get-started.md

raphaelchaula:
running 'npx -p @storybook/cli sb init' alone has some issues to some, it won't install storybook dependencies at all, I fixed it by installing storybook cli then run sb init, hence I suggested changes to docs.

### DIFF
--- a/content/intro-to-storybook/react/en/get-started.md
+++ b/content/intro-to-storybook/react/en/get-started.md
@@ -23,6 +23,16 @@ cd taskbox
 npx -p @storybook/cli sb init
 ```
 
+In case storybook won't install dependencies
+
+```bash
+# Install storybook cli in Dev mode
+yarn add -D @storybook/cli
+
+# initialize sb
+yarn sb init
+```
+
 <div class="aside">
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@raphaelchaula thank you for the time and effort you put into the pull request. Hope you don't mind but this is not going to be  merged. Reason being is that we want to keep the tutorial lean and concise. And on point. Including these changes would add a bit of entropy to the get started section and we want you and the other readers to well get started and have the environment setup with the minimum amount of effort, so that you can start working on the components/stories as fast as you can.

I'm  truly sorry that you've experienced issues with Storybook and what i would like from your end is that you create a issue in the Storybook [repository](https://github.com/storybookjs/storybook), detailing what you've experienced and if possible a small reproduction to help the maintainers triage the issue quicker and possibly offer you a solution. Sounds good?

Stay safe.

raphaelchaula:
Sounds good,
I will surely open the issue and suggest the changes.

jonniebigodes:
Thank you. The maintainers would appreciate it :)

PR #402, opened 2020-10-23
Update English design systems guide to reflect the fact new GH repos now have a default branch named 'main'

itsojon:
First off, thank you for a wonderful tool in Storybook and for this superb guide!

As described [here](https://github.com/github/renaming#new-repositories-use-main-as-default-branch-name), new GitHub repos now have a default branch named `main` instead of `master`.

I updated the code snippets in both the English version of the Design Systems guide and the Portuguese version, and other references to "master" in the English version (I don't speak Portuguese).

When "master" is referenced in copy (not code), I replaced it with "the default branch" since I figure it will take a little time for people to get used to references to `main` like "we’ve been pushing to main".

Remaining TODOs (I'm happy to take a stab at these if you like, except for non-English copy changes):
- Update code in https://github.com/chromaui/learnstorybook-design-system that references "master", like [this code](https://github.com/chromaui/learnstorybook-design-system/blob/2d0450a27643675ecd93b3d18656705c703c8bca/.github/workflows/push.yml#L7)
- Update the other guides
- Update the Portuguese copy for the Design Systems guide

### DIFF
--- a/content/design-systems-for-developers/react/en/architecture.md
+++ b/content/design-systems-for-developers/react/en/architecture.md
@@ -63,9 +63,9 @@ Then follow GitHub's instructions to setup the (the so-far mostly empty) reposit
 git init
 git add .
 git commit -m "first commit"
-git branch -M master
+git branch -M main
 git remote add origin https://github.com/your-username/learnstorybook-design-system.git
-git push -u origin master
+git push -u origin main
 ```
 
 Be sure to replace `your-username` with your own account name.
--- a/content/design-systems-for-developers/react/en/distribute.md
+++ b/content/design-systems-for-developers/react/en/distribute.md
@@ -167,7 +167,7 @@ GH_TOKEN=<value you just got from GitHub>
 NPM_TOKEN=<value you just got from npm>
 ```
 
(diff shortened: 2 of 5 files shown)


itsojon:
Realized a default branch name of `main` also broke automatic publishing via `auto shipit` since that command assumes a default branch name of `master`, so I updated that to `auto shipit --base-branch=main`.

jonniebigodes:
@itsojon Thank you very much for your help on this. I'm going to take a closer look at this in a bit. If everything checks out i'm going to merge this. Regarding the translation for this guide. Indeed it needs a bit of attention as it's content is currently not matching the original version. I have yet to to update it. If you could update the other tutorial regarding this item it would be fantastic and we would really appreciate it.

Stay safe

jonniebigodes:
@itsojon i've checked and we're in good shape. I'll merge it soon. Thank you for the time you put into this pull request. 

Stay safe

itsojon:
@jonniebigodes you're very welcome! I'll make time to work on the other guides in the coming days. I've created this issue to track the subtasks: https://github.com/chromaui/learnstorybook.com/issues/408

PR #422, opened 2020-12-13
Typo in creating-addons.md (English version)

luisabellan:
it's >> its
possesive adjective


jonniebigodes:
@luisabellan hope you're doing ok during these times. I'm going to close this as the create addons section was moved into its own tutorial. And with that we'd love a translation. If you and @IvanSotelo could synch up and work on it we could merge it in a flash. 

Let me know 

Stay safe

Issue #431, opened 2021-01-06
Intro to Storybook templates


jonniebigodes:
Looking at the present state of the Intro to Storybook in its various forms (frameworks), there's a lot of work that the reader needs to do in order to get a minimal viable project setup with Storybook and its framework of choice. As the tutorial is an introductory one the reader should be able to get onboarded with a minimum effort on his part. What i'm thinking here is the following:
- Create a template for the various frameworks currently documented (React, Angular, Vue, Ember, React Native).
- Said template would streamline the overall experience and reduce the amount of steps that the reader has to go through to get onboarded.

With this i would love to hear from @dannyhw as you're "leading the charge" (pardon the bad pun) on the React Native version of Storybook and from @gossi as you're in the forefront of the ember experience with Storybook.

Let me know what you think.

Stay safe

gossi:
Can say for ember, when you do `ember install @storybook/ember-cli-storybook` this will run the `sb init`. Not the real one, ember-cli has the concept of blueprints instead, which mimics what `sb init` is doing here. It creates the config and hooks into ember (or should do it). I think it would even be better if it would execute `sb init` (less mainteneance for sure, higher consistency)

This is what I see as an entry point for ember. What do you have in mind this _template_ is doing?

dannyhw:
Speaking from my experience with the react native project the part which benefits most from having a template is the setup of things like CSS, fonts, icons etc which depending on the framework can be kind of complicated. In the react native example the first section is/was a series of copy and paste for setting up things that aren't directly related to the content of the tutorial such as setting up the font as mentioned. Having this as part of a template is beneficial for reducing setup time and potential for error. The cost is one of maintaining the template so it does not become out of date. This could be as simple as a repo containing the project ready for the tutorial.

The other option would be to simplify the tutorial project so it doesn't include additional files or setup etc.

jonniebigodes:
The way i'm thinking on putting this into effect is basically a 2 step approach. I want to start small and create the repo templates for the various frameworks so that when the reader lands on the get started section for it's framework of choice he/she would only need to run something like:

```shell
# clone repo
npx degit chromaui/intro-storybook-react-template taskbox

cd taskbox

# install dependencies
npm/yarn install
```

And he or she would be good to go and move onto the next section. I'm well aware of the cost associated with maintenance but it can be mitigated with some tooling already in place, for instance [renovatebot](https://github.com/renovatebot/renovate). 

[...]

PR #433, opened 2021-01-07
change destination of assets & css in Svelte Intro

rodgco:
In `Get Started` for Svelte the existing CSS and assets (fonts & icons) where saved at `./src`. It makes more sense to make those changes directly in `./public`.


jonniebigodes:
@rodgco thank you for the time and effort you put into the pull request, but sadly i'm going to close it. And go into a little bit of detail on why.

I'm well aware that you can put the assets directly into the `public` and you'll get access to them. But with a recent change introduced by CRA and React 17 some compromises had to be made for the css and assets (icons and fonts) to be properly displayed in the tutorial. Moving them into the `src/assets` folder allowed a more universal approach and get you and the rest of the community unblocked and able to go through the through the tutorial without adding more and more steps to the get started section, when the goal of that section is to get you onboarded with the minimum effort and move onto creating the components and stories. This approach serves a bit of a middle ground before #431 is set into motion. I'm currently researching this and trying to coordinate with the community to get a good starting point so that when you or any other reader comes into the tutorial he/she can go though the tutorial more smoothly. If you're able i would like to hear your thoughts in the issue i've linked.

Stay safe

rodgco:
@jonniebigodes the direction you're taking makes perfect sense. I'm new to SB and to Svelte, I'll try to contribute as much as I can.

PR #492, opened 2021-05-12
Update build.md

snilan:
The Button story has no actions defined and the screen grab is actually showing the Link story. It's a bit confusing. Also the excerpted example code for the buttonWrapper story doesn't match up with the codebase. I wasn't sure if the tutorial wanted us to add the action to the Button story or meant to show us the action in the Link story. I took the former approach but the video should be updated


jonniebigodes:
Going to close this pull request as the tutorial was updated to factor in these changes.

PR #512, opened 2021-06-18
Indicate to the user the need to create preview-head.html

MichaelArestad:
I updated the instructions to indicate the need to create `.storybook/preview-head.html`. It could maybe be a bit more explicit, but I think this change could do the trick.


jonniebigodes:
Thanks for the fix @MichaelArestad i'm good with it, merging.

PR #529, opened 2021-08-31
Update simple-component.md

DohertyC12:
https://github.com/storybookjs/react-native/issues/58
Storybook does not load the Task component here - just the Storybook framework
You should run storybook server and then after run the react native app then if you refresh the webpage for react native server the stories should load.

### DIFF
--- a/content/intro-to-storybook/react-native/en/simple-component.md
+++ b/content/intro-to-storybook/react-native/en/simple-component.md
@@ -168,6 +168,7 @@ export default function Task({ task: { id, title, state }, onArchiveTask, onPinT
 ```
 
 The additional markup from above combined with the styling we created earlier yields the following UI:
+(Does not create the Storybook here, keep going - https://github.com/storybookjs/react-native/issues/58)
 
 <video autoPlay muted playsInline loop>
   <source


content/intro-to-storybook/react-native/en/simple-component.md
```diff
@@ -168,6 +168,7 @@ export default function Task({ task: { id, title, state }, onArchiveTask, onPinT
 ```
 
 The additional markup from above combined with the styling we created earlier yields the following UI:
+(Does not create the Storybook here, keep going - https://github.com/storybookjs/react-native/issues/58)
```
jonniebigodes, line 172: @DohertyC12 instead of putting this sentence here as it may mislead the readers. We have room for improvement. We can probably add the following below the video:

```markdown
<div class="aside"
💡 <strong>Note:</strong> Don't forget to start your Storybook server before running your application to see the changes in effect.
</div>
```

The reason being is that throwing in that sentence might mislead users and break the flow of the section. As one might read that everything should be working and out of the box you're telling them not to create Storybook and point them to an issue. See where I'm going with this?

jonniebigodes (changes requested):
@DohertyC12 thanks for the time and effort you've put into this pull request. We appreciate it. I left one item for you to address if you don't mind. Let me know and we'll go from there.

Stay safe

PR #560, opened 2021-12-18
Missing imports

awareness481:
`fireEvent` & `within` are used in the code examples but are not imported.


jonniebigodes:
@awareness481 thanks for taking the time and effort you've put into this pull request and help us catching those missing imports. We appreciate it. I've checked and it looks good on my end. Going to merge this,

Happy Hollidays!

Stay safe

PR #563, opened 2022-01-03
Suggested gotcha note for the English version

cabbiepete:
Needs translation.

### DIFF
--- a/content/visual-testing-handbook/react/en/automate.md
+++ b/content/visual-testing-handbook/react/en/automate.md
@@ -74,6 +74,8 @@ yarn chromatic --project-token=<project-token>
 
 ![Chromatic running](/intro-to-storybook/chromatic-manual-storybook-console-log.png)
 
+**GOTCHA: If you're project has less than 2 commits chromatic publishing will fail while `Retrieving git information` add another commit and try again.**
+
 With this one command, you published your Storybook, triggered Chromatic to capture an image snapshot of each story (in a standardized cloud browser), and set the snapshot as the baseline.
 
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@cabbiepete thanks for taking the time and effort to create the pull request and improve the tutorial. But looking at it actually is not actually necessary. To give you a bit of context, if you follow along with the tutorial you should have 2 commits when publishing to Chromatic. One when you set up the repo in GitHub and the other one after adding the Chromatic CLI to your project. It was implemented that way to avoid situations like the one you're mentioning in this pull request.

Hope you don't mind but in this case, I'm closing the pull request.

Have a great day!

Stay safe

cabbiepete:
> @cabbiepete thanks for taking the time and effort to create the pull request and improve the tutorial. But looking at it actually is not actually necessary. To give you a bit of context, if you follow along with the tutorial you should have 2 commits when publishing to Chromatic. One when you set up the repo in GitHub and the other one after adding the Chromatic CLI to your project. It was implemented that way to avoid situations like the one you're mentioning in this pull request.
> 
> Hope you don't mind but in this case, I'm closing the pull request.
> 
> Have a great day!
> 
> Stay safe

Sure your call ultimately. 

Perhaps it's worth simplifying the cli explaination? That seems to assume you did a shallow clone when actually this other situation also clearly exists, most developers won't even know what a shallow clone is as it is not a day to day use case for git.

jonniebigodes:
@cabbiepete we can work with that if you're ok with opening up an issue in the CLI's [repo](https://github.com/chromaui/chromatic-cli) and mention your latest comment the team can take a look at it and adjust the messaging for clarity. Sounds good?

PR #572, opened 2022-01-31
Update get-started.md

lrcorona2014:


jonniebigodes:
@lrcorona2014 hope you don't mind, but I'm going to close this pull request, the reason being is that we much rather have a full update of the Spanish translation,  instead of just adding this small change? The reason I'm asking this is due to the fact that those translations are a bit outdated and require some work to match the current version of the tutorials. If you're ok with it. Drop a message in the issue set up for translations and hopefully connect with other community members that have expressed interest in translating in the past.


Hope you have a great weekend 

Stay safe

PR #587, opened 2022-02-17
Grammar updates across all content

benkane:
Change setup to set up when it's used as a verb


kylegach (approved):
Thanks once more, Ben! So helpful that you went the extra mile and did the fix across the entire repo. 🙏 

Great first contribution, too! ❤️  🥳

PR #597, opened 2022-05-03
Fix task ids in mapped data

murphyslaw:
The id of the task was not interpolated but passed as a literal `${task.id}` string, which meant that all tasks had the same id, which in turn prevented the interactions from working properly later on.

### DIFF
--- a/content/intro-to-storybook/react/en/screen.md
+++ b/content/intro-to-storybook/react/en/screen.md
@@ -47,7 +47,7 @@ const TaskBoxData = {
 +   );
 +   const data = await response.json();
 +   const result = data.map((task) => ({
-+     id: `${task.id}`,
++     id: task.id,
 +     title: task.title,
 +     state: task.completed ? 'TASK_ARCHIVED' : 'TASK_INBOX',
 +   }));


murphyslaw:
@jonniebigodes I tested around further and figured out, that I made a stupid typo mistake. Instead of `${task.id}` I actually wrote `$(task.id)`, which did not evaluate the variable but just simply set the ID to literally `$(task.id)`. This meant, that every task ID was assigned the string `$(task.id)`, which then was the reason for my interactions to fail, because they relied on each task having different ids.

So the fix on my side should have been to use curly braces instead of round ones and leave the type as string. I went too far with the change I proposed here in the pull request, which really changed the type to an integer. Even though this helped to solve my issue with assigning correct, unique IDs, it was not needed if the variable would just have been evaluated correctly with the curly braces and turned into a string. Good that you had type checkings included, so my solution was exposed as a too far reaching solution.

Again, I'm really sorry for spending your valuable time because of my typing mistake.

jonniebigodes:
@murphyslaw no problem whatsoever, glad that I was able to help you triage and point you in the right direction and helped fix the issue you had. Also no need to apologize whatsoever, we're all learning here :)!

Hop you have a great weekend.

Stay safe

Issue #603, opened 2022-05-17
Import library name is outdated


smilee:
<img width="845" alt="image" src="https://user-images.githubusercontent.com/5793796/168928661-7ba036a4-e485-4059-a3a7-436a80393837.png">

@storybook/api should be @storybook/addons

smilee:
https://github.com/chromaui/learnstorybook.com/pull/604

jonniebigodes:
@smilee thanks for taking the time to open up the issue and pr. Sadly I had to close and will do the same here. Hope you understand but the change you proposed would lead to an issue with the tutorial. 

Hope you have a great weekend.

stay safe

PR #604, opened 2022-05-17
Update import library name

smilee:
Issue: https://github.com/chromaui/learnstorybook.com/issues/603

### DIFF
--- a/content/create-an-addon/react/en/track-state.md
+++ b/content/create-an-addon/react/en/track-state.md
@@ -30,7 +30,7 @@ The Addon Kit preconfigures the `Tool` to use globals. Let’s rename the global
 
 ```diff:title=src/Tool.js
 import React, { useCallback } from 'react';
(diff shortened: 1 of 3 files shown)


jonniebigodes:
@smilee thanks for taking the time and effort with this pull request. Sadly, this is not being merged as this would introduce an issue, breaking the tutorial at this stage. To give you a bit of context if you or any other user made the change you're proposing he/she would get the following while building the addon:
blankscreen-storybook
blankscreen-devtools

The reason being is that the import you're adding is used/affects one part of Storybook, not the part that this section references. 

I'm aware that the callout pointing to `@storybook/addon` could be a bit misleading and I'm going to follow up with the tutorial's author and clarify the wording around it so that our readers don't get confused as you probably did.

With that out of the way, going to close this pull request.

Hope you have a great weekend!

Stay safe

PR #626, opened 2022-07-31
Commit store actions in PureInboxScreen stories

tsega:
Unless the store actions (**ARCHIVE_TASK** and **PIN_TASK**) in `PureInboxScreen.storeis.js` are committed to the store, the archiving and pinning of tasks in Storybook does not produce any visual results.

### DIFF
--- a/content/intro-to-storybook/vue/en/screen.md
+++ b/content/intro-to-storybook/vue/en/screen.md
@@ -191,11 +191,13 @@ import PureInboxScreen from './PureInboxScreen.vue';
 +     },
 +    },
 +   actions: {
-+     pinTask(context, id) {
++     pinTask({ commit }, id) {
 +       action('pin-task')(id);
++       commit('PIN_TASK', id);
 +     },
-+     archiveTask(context, id) {
++     archiveTask({ commit }, id) {
 +       action('archive-task')(id);
++       commit('ARCHIVE_TASK', id);
 +     },
 +   },
 + });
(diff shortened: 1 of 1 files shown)


jonniebigodes:
@tsega thanks for taking the time and effort into this pull request and willingness to help improve the tutorial. We appreciate it. But sadly, this will be closed as the change you're introducing will result in an issue later in the tutorial. To give a bit of context on this, when you apply this change and update the code later on for the Interaction testing with the `play` function, it will only be able to run once as the task is already pinned and Storybook is generated a bit of a "false positive", which could potentially mislead our readers. To avoid that, I went with this approach instead, I know that's a bit off, but with it, we're able to demonstrate the intended patterns, supply a provider and successfully run an interaction test with the play function. Hopefully, sometime in the near future, the Vue3 implementation for Storybook will be updated and enable a better workflow and avoid this known issue, and with it, apply the changes you're proposing here without breaking other parts of Storybook.

I hope I have clarified you a bit on this, and I sincerely hope you have a great week.

Looking forward to the next contribution.

Stay safe.

tsega:
Thanks for the detailed explanation @jonniebigodes, I will move on to the next part of the tutorial and see if I can get my head around it. 

I really appreciate that you looked into this. I look forward to be of any help.

jonniebigodes:
@tsega, no need to thank whatsoever. It was my pleasure, and I am sorry for that part of the tutorial not being clear enough and a bit misleading. As I've mentioned, hopefully, that part of Storybook will be updated sometime in the near future, and we're able to provide you and the rest of the community with a similar experience as with other frameworks. 

Feel free to reach out to us by opening an issue if you encounter an issue with the tutorials. I'll be more than glad to follow up with you and hopefully get you unblocked.

PR #633, opened 2022-08-24
Create copy button for code blocks.

### LINKED ISSUE #585: Copy-paste is difficult with diff code blocks
After working my way through the tutorials for a while now, I wanted to give some feedback regarding code blocks with diff sections. For example:

image

As you might imagine, it gets a bit annoying to have to delete the `+` on every line after copy and pasting. I've seen some other sites that take a different approach:

- [React docs](https://reactjs.org/docs/components-and-props.html#composing-components) - The changes have a different background color.
- [Gatbsy Docs](https://www.gatsbyjs.com/docs/tutorial/part-2/) - Different background color **and** a button to copy the code block.

Perhaps it would be better without the `+` signs?

**UPDATE**

Just FYI, after a little digging it looks like the styling comes from the [@storybook/design-system](https://github.com/storybookjs/design-system/blob/master/src/components/Highlight.tsx) dependency. I suppose it would need to change there?

jeffslofish:
I would consider this a rough draft to start a conversation, but it solves the issues of #585. 

For diff blocks, it copies what the final code should look like, so it removes lines starting with - and removes the + from lines starting with + and leaves the other lines as they are.
Based heaving upon https://github.com/iamskok/gatsby-remark-code-buttons and some ideas from https://github.com/gatsbyjs/gatsby/pull/15834/files#diff-112c73a49a03da844b5d0316314e16e7bbfd78528f7d31e54305f4dd523c50c0

Closes #585

https://user-images.githubusercontent.com/1240484/186472469-809def20-3715-4e89-8046-8817cbc8f57b.mp4


jonniebigodes:
@jeffslofish Wow, this is amazing. Thank you very much for taking the time and helping us out with this. We're truly thankful for it. I've checked, and looks really great, but if you wouldn't mind holding on a bit more before this gets merged in, as we're in the middle of an update. Once the update is released, I'll let you know so that we can coordinate with you and get this updated and merged in and close the issue.

Sounds good?

Looking forward to hearing from you.

Stay safe

jonniebigodes:
@jeffslofish just merged it in. Thanks for taking care of this issue and helping us provide a better way for users to follow the tutorial. As a follow-up to this pull request, I wanted to ask you if you could message me (same username) on our [Discord server](https://discord.gg/storybook) ?

Hope you have a great weekend.

Stay safe

PR #639, opened 2022-09-04
Removed styling only addressed in topic 'Addons'

H3AR7B3A7:
In the topic 'Addons' this component is tested by using the Controls Addon. At this point the overflow behavior with long strings is addressed by adding the styling and a visual regression test. Since this style is already added at the start of the course, someone following along the tutorial will not see the text overflow the component like in the examples. So it should not be there yet at this point.


jonniebigodes:
@H3AR7B3A7, thanks for taking the time and effort you've put into this pull request and helping us improve the tutorial by catching that small but rather important typo. I've checked, and it's good on my end. The styling should not be referenced here. Sorry for the confusion, but that information slipped through when the tutorial was being updated earlier.

Going to merge this.

Hope you have a great rest of the weekend.


Stay safe

PR #642, opened 2022-09-12
More detail on installing Degit

sosensible:
I was not sure if this was something that was installed or not on its own. Then today, after speaking on this, the same question is coming from others.

### DIFF
--- a/content/intro-to-storybook/vue/en/get-started.md
+++ b/content/intro-to-storybook/vue/en/get-started.md
@@ -11,7 +11,7 @@ Storybook runs alongside your app in development mode. It helps you build UI com
 
 ## Set up Vue Storybook
 
-We'll need to follow a few steps to get the build process set up in our environment. To start with, we want to use [degit](https://github.com/Rich-Harris/degit) to set up our build system. Using this package, you can download "templates" (partially built applications with some default configuration) to help you fast track your development workflow.
+We'll need to follow a few steps to get the build process set up in our environment. To start with, we want to use [degit](https://github.com/Rich-Harris/degit) to set up our build system. Using this package, you can download "templates" (partially built applications with some default configuration) to help you fast track your development workflow. (If you don't have [degit](https://github.com/Rich-Harris/degit) installed, follow the link to install it.)
 
 Let’s run the following commands:
 


jonniebigodes:
@sosensible, thanks for taking the time and effort to create this pull request and helping us improve our tutorials. To give you a bit of context, running `npx somepackage` does not install the package but instead downloads it, caches it, and executes it. You can read more about it [here](https://nodejs.dev/en/learn/the-npx-nodejs-package-runner/). We went with this approach to avoid having our readers install another package just to get started with the tutorial, or having them go through the process of forking the template repo, cloning it, and running it on their machines and inadvertently having them push a pr against the template repo owned by Chromatic which could cause some confusion. 

One thing that we could do is clarify what the command actually does, and should the reader wants to install it is up to them.

What do you think? Sounds good to you?

Hope you have a great week.

Stay safe

sosensible:
I think we need fresh eyes to answer the question, first pass users. I
don't have fresh ones ATM... but will stay alert as more jump in.

Also, I will be submitting some fixes to the VUEX implementation. I had to
debug and fix that when I went through the turorial.

*John Farrar*
*SOSApps*
*Servant Oriented Software*

jonniebigodes:
Totally agree with you on the fresh eyes. And please do. Looking forward to seeing the changes to the data/screen sections where VueX comes into play. Before committing to them, are you ok with opening up an issue detailing it a bit more so that we can continue the discussion there and polish the changes so that once it's committed, it's a simple click of a button to merge them in

Looking forward to reading it.

jonniebigodes:
@sosensible, if you're ok with opening up an issue with the feedback collected here also, I wanted to let you know that we've moved away from VueX as the data provider library for Pinia as per Vue recommendations. The tutorial was updated accordingly, and please let me know if you spot any issues.

Looking forward to hearing from you.

Hope you have a great weekend.

Stay safe

jonniebigodes:
Closing this pr for now. Let me know if you still wish to work on this, and I'll gladly follow up with you on our issue tracker until we find a suitable solution that helps you go through the tutorial without any issues.


Hope you have a fantastic weekend.

Stay safe

PR #658, opened 2022-10-16
Fix the shell script to install a11y addon

takuyakikuchi:
## Issue

The shell script to install a11y addon is not correct.

## Changes

The shell script updated.


jonniebigodes:
@takuyakikuchi, thanks for this. It seems that a recent change in the tutorial didn't actually cover it. Appreciate it you taking the time to patch this up. On my, this is good to go. Going to merge it. 

Hope you have a great week.

Stay safe

PR #659, opened 2022-10-20
Update screen.md

timur-svoboda:


jonniebigodes:
@timur-svoboda, thanks for taking the time and effort into this pull request and helping improve our tutorials. We appreciate it. But in this case, the wording is actually correct. To give you a bit of context, the choice of "change" doesn't mean only story it could also be a change in the component implementation as documented extensively in the tutorial. With that, sadly, I'll have to close this pull request.

Hope you don't take it the wrong way. I hope you have a great week.

Stay safe

Issue #662, opened 2022-10-23
Grammatical mistakes in Spanish translation


shogohida:
I found 4 grammatical mistakes in Spanish translation. "componente simples" should be "componente simple" because the noun is singular and the adjective also must be singular.

https://github.com/chromaui/learnstorybook.com/blob/master/content/intro-to-storybook/vue/es/simple-component.md
https://github.com/chromaui/learnstorybook.com/blob/master/content/intro-to-storybook/svelte/es/simple-component.md

jonniebigodes:
Closing this per comment in the associated pull request.

PR #664, opened 2022-10-23
Fix plural adjectives to singular in Spanish translation

shogohida:
I made this pull request to solve [this issue](https://github.com/chromaui/learnstorybook.com/issues/662)

### DIFF
--- a/content/intro-to-storybook/svelte/es/simple-component.md
+++ b/content/intro-to-storybook/svelte/es/simple-component.md
@@ -1,6 +1,6 @@
 ---
-title: 'Construye un componente simples'
-tocTitle: 'Componente Simples'
+title: 'Construye un componente simple'
+tocTitle: 'Componente Simple'
 description: 'Construye un componente simple en aislamiento'
 ---
 
--- a/content/intro-to-storybook/vue/es/simple-component.md
+++ b/content/intro-to-storybook/vue/es/simple-component.md
@@ -1,6 +1,6 @@
 ---
-title: 'Construye un componente simples'
-tocTitle: 'Componente Simples'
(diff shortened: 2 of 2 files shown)


jonniebigodes:
@shogohida sorry for the delayed response on this, as I was tasked with other documentation items. Once again, sorry for this, but sadly this pull request will also be closed. Even though the word exists in the Spanish dictionary and could be used, in this case, it's actually inaccurate as the change you're introducing would be a grammar error. For instance, a case where the word would be used would be the following:

"¿Como puede alguien, siendo un simple mortal, puede retar a Dios?", which translates to "How can anyone, being a mere mortal, ever challenge God?" . Here it's accurate and it's the correct usage. But the case you're using it's not and the word should also be pluralized. To give you a bit more context, I've reached out to both the community for some insights and a close friend of mine which is a native Spanish speaker/writter. 

Have a great weekend

Stay safe

shogohida:
@jonniebigodes 
Thanks for your explanation! To be honest, I don't understand yet why both of "un componente simple" and "un componente simples" exist and their differences depending on the contexts... 

However, you've reached out to several people so your choice should be respected 👍 I'm not a native speaker of Spanish or Portuguese in the end hahaha

Hope to contribute to other translations in the future!

jonniebigodes:
@shogohida no need to thank whatsoever. Glad that I was able to provide you with some context. And it's one of those nuances that some Latin-based languages (e.g., Portuguese, Spanish) have that can elude non-native speakers. You might think that you're fixing a grammar error, such as in this case, but you're introducing one. 

And please feel free to do so; we'd welcome updates to the translations.

PR #668, opened 2022-10-24
Fix EditTask play function

timur-svoboda:
If you do not put await in front of userEvent, then when moving through this step in the "Interactions" tab in the Storybook, there will be an instant transition to the next step, due to which the value of the input field will not have time to change, that is, the test will fail.

Video: https://monosnap.com/file/jeGnfLi5oysTYK5QOdxcu5rLq7SAqN


jonniebigodes:
@timur-svoboda, thanks for taking the time and effort into this pull request and your willingness to help us improve the tutorial by finding this small but rather crucial typo. We appreciate it 🙏 ! I checked it, and indeed you're correct, as the interaction test will break as currently documented if you, for example, go back to the beginning and start iterating them step by step, as you have shown in the video.

Going to merge this.

Hope you have a great week.

Stay safe

PR #698, opened 2023-03-31
Adding node version in workflow

GreenbowAllan:


jonniebigodes:
@GreenbowAllan thanks for taking the time to put together this pull request but I hope you don't take it the wrong way as the tutorial required more changes to be added as part of the Storybook 7 release. A pull request is already up and should be published shortly.

Have a great week.

Stay safe

PR #727, opened 2023-08-23
Update deploy.md

subasically:


jonniebigodes:
@subasically appreciate the time you took to put together this pull request, but before moving on, could you provide some explanation on why you're updating the command? The reason I'm asking is mainly that the documentation is accurate as it refers to the script that is added to the package.json file when Storybook runs the install process, which is already provided as part of the template.


Let me know, and we'll go fro there.

Have a fantastic day.

Stay safe

jonniebigodes:
@subasically going to close this for now. Feel free to follow up with me so that we can continue to work through it and help you and the community improve the tutorial.

Have a fantastic weekend.

PR #745, opened 2023-12-12
Add plausible analytics

Integrayshaun:
## What changed?
- [x] Add Plausible plugin with patch for Netlify proxying
- [x] Add redirects for Netlify proxying
- [x] Add `patch-package` for patching plugin
- [x] Alphabetize package scripts (Because I love us and want to be nice to us)

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,4 +1,5 @@
 const isDeployPreview = process.env.CONTEXT === 'deploy-preview';
+const isProduction = process.env.CONTEXT === 'production';
 const permalinkBase = isDeployPreview ? process.env.DEPLOY_PRIME_URL : 'https://storybook.js.org';
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


gatsby-config.js
```diff
@@ -111,6 +111,16 @@ module.exports = {
       },
     },
     `gatsby-plugin-sitemap`,
+    ...(!isDeployPreview
```
kylegach, line 114: This would result in the plugin running during local development. Will Plausible have any issue with that?

kylegach (approved):
👍

gatsby-config.js
jonniebigodes: I ran it locally, and checking the dev tools, nothing is being shown regarding Plausible, so it seems we're in the clear. @Integrayshaun can you confirm if anything is being tracked

Integrayshaun: Plausible wont, but I'm sure that we don't want localhost records in our analytics. Fixed

PR #749, opened 2024-01-02
Add id prop to input element in simple-component.md

dangkhoipro:
The value of htmlFor is id (not the input's name), so you have to put the same value into the id of the input. Btw your current example is still working because you are putting the input inside of the label, otherwise, it will not work as expected


jonniebigodes:
@dangkhoipro thanks for taking the time to put together this pull request and helping us improve the tutorial. We appreciate it 🙏 ! I've checked and all is good on my end. 

Sorry for the delayed response, but I was otherwise engaged with some other documentation issues.

Hope you have a fantastic day.

Stay safe

PR #789, opened 2024-12-17
fix analytics

winkerVSbecks:


kylegach:
How do I verify this?

https://deploy-preview-789--storybook-tutorials.netlify.app/proxy/api/event

Looks correct?

winkerVSbecks:
> How do I verify this?
> 
> https://deploy-preview-789--storybook-tutorials.netlify.app/proxy/api/event
> 
> Looks correct?

Yea, that's the right page. It's an POST endpoint so if you visit it then you see that.

kylegach (approved):
Thank you so much for this!

PR #805, opened 2025-05-30
fix: update import path for fn

shiva-raghav:
fix: update import path for fn to use storybook/test instead of @storybook/test as per migration guide for Storybook 9.0


jonniebigodes:
@shiva-raghav, thanks for taking the time to put together this pull request and helping us improve the tutorial. We appreciate it 🙏 ! Sadly, it's going to be closed as adding this small change will only affect Storybook 9, which is coming out soon. With that, it introduces some functional issues as the rest of the tutorial is not yet updated (including templates). 

Have a great day

Stay safe

